PR #26, opened 2018-03-15
Use more C variables in SART inner loops

jacklovell:
There were a couple of instances in the SART inner loops where Python
variables were used, either because the variable hadn't been declared
as a C variable or because the Python version rather than the C version
of the variable was used.

Using the C variables instead results in an almost 5-fold speed increase
on a test of inverting 32 LOS measurements on 900 cells, with no discernable
change in output.


mattngc:
Cool, nicely spotted. I hadn't done any profiling on that yet as it was still a draft of the algorithm. But I guess its close enough to final now.

PR #28, opened 2018-03-26
Modify definition of beta_laplace to match Carr RSI 2018 definition

jacklovell:
modified:   cherab/tools/inversions/sart.pyx


CnlPepper:
Just wanted to check, in changing the divide to a multiply shouldn't the constant change to 0.1 from 10 rather than 0.001.

jacklovell:
If you were to exactly preserve the functionality, yes. But according to Matt 0.01 is a more useful default (the original default value was a hangover from early in the implementation of this SART inversion, as I understand).

The 0.001 was a typo: it should have been 0.01. I've updated the branch now with the new value, but double check with Matt that this is really correct before merging.

mattngc:
Jack is right. The new default value is better based on the RSI paper.

PR #35, opened 2018-06-11
Add ability to specify arbitrary Tikhonov matrix for NNLS

jacklovell:
By default, the identity matrix multiplied by alpha is used in the
regularised NNLS (L2 regularisation). However, other constraints
such as smoothness (isotropic or anisotropic) may be desired, which
require a different Tikhonov regularisation matrix. This commit allows
an end user to specify such a matrix.

### DIFF
--- a/cherab/tools/inversions/nnls.py
+++ b/cherab/tools/inversions/nnls.py
@@ -19,18 +19,31 @@ import numpy as np
 import scipy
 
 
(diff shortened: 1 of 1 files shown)


mattngc:
This looks good to merge. But in future I would consider a refactoring of the function name. In reality we should just have a library called Tikhonov with a number of different functions and smoothing operators to choose from.

PR #40, opened 2018-09-05
Fix orientation of bolometer slits and foils

jacklovell:
The rotation transform should be such that the normal vector for
slits and foils points in the local z direction, and the x and y
basis vectors point in the local x and y directions respectively.
This is important when the elements are rectangular rather than
square.

Previously, the arguments to the rotate_basis function for these
elements resulted in the normal vector pointing in the local z
direction (correct), but the basis_x vector pointing in the local
y direction (a rotation of 90 degrees from what it should be).
This commit updates the arguments to rotate_basis, so that it is
the basis_y vector which points in the local y direction.


mattngc:
I agree that this is a sensible change.

PR #41, opened 2018-09-05
WIP:Support bolometer slits and foils with rounded edges

jacklovell:
The most common design of bolometer foil currently in use is not
truly rectangular, but has rounded edges with a curvature radius
of 0.5mm. This reduces the area of the foil by a little under 5%,
with corresponding effects on the sensitivity and etendue of the
foil. These become a source of systematic error if the foil is
treated as a simple rectangular detector.

This commit enables treating the foils as rounded rectangles.
Because the TargettedPixel is necessarily rectangular, the rounding
is achieved by creating a mask just above the pixel, which covers
the corners of the pixel. The remaining un-covered area of the pixel
is in the shape of the rounded rectangle.

The mask is created programatically, which enables the radius of
curvature of the foils to be specified at instantiation. This allows
other bolometer sensor designs to be used (earlier bolometer sensors
from the same manufacturer for example had a radius of curvature of
0.2mm, and other sensor designs do not have curved corners at all).
By default, the sensors are treated as rectangles with 0 curvature
radius, which maintains the previous behaviour.

It is also possible to specify a radius of curvature for the slits.
Although no systems that I'm currently aware of actually use rounded
slits, this does improve the flexbility of the code to deal with
future designs of slit, for very little additional work.


jacklovell:
Before merging, I'd like to check whether you think this is the best approach for this problem, or whether there's anything better. I also haven't fully tested this yet (e.g. by calculating étendues), so have marked the PR as a work in progress until then.

cherab/tools/observers/bolometry.py
```diff
@@ -127,14 +128,16 @@ def observe(self):
...
+    def __init__(self, slit_id, centre_point, basis_x, dx, basis_y, dy, dz=0.001,
+                 parent=None, csg_aperture=False, curvature_radius=0):
 
         self._centre_point = centre_point
         self._basis_x = basis_x.normalise()
         self.dx = dx
         self._basis_y = basis_y.normalise()
         self.dy = dy
         self.dz = dz
+        self._curvature_radius = curvature_radius
```
mattngc: You don't have any validation on the curvature_radius variable in the BolometerSlit. But you do have it already for BolometerFoil. Please copy over.

PR #43, opened 2018-09-19
Feature/voxels

mattngc:
Major refactor of the bolometry and inversion grid classes.

- Added two new voxel classes, allowing cylindrically symmetric voxels to have an arbitrary polygon cross-section.
- Reduced the number of Bolometer file formats. Users will now choose there own formats in their application and instantiate the classes directly.
- Tidied up the way transforms are handled.
- The BolometerFoil class is now a sub class of the TargettedPixel observer.
- calculate_sensitivity() now returns a sensitivity vector and doesn't store the results on the class. This change now supports calculations on emission sources in parallel resulting in a substantial speed up if reflection effects are not being considered.
- Many minor improvements, the code is now much cleaner.

Assigning the merge review to @jacklovell.

### DIFF
--- a/cherab/tools/inversions/voxels.pyx
+++ b/cherab/tools/inversions/voxels.pyx
@@ -1,0 +1,285 @@
+
+# Copyright 2016-2018 Euratom
+# Copyright 2016-2018 United Kingdom Atomic Energy Authority
+# Copyright 2016-2018 Centro de Investigaciones Energéticas, Medioambientales y Tecnológicas
+#
+# Licensed under the EUPL, Version 1.1 or – as soon they will be approved by the
+# European Commission - subsequent versions of the EUPL (the "Licence");
+# You may not use this work except in compliance with the Licence.
+# You may obtain a copy of the Licence at:
+#
+# https://joinup.ec.europa.eu/software/page/eupl5
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the Licence is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied.
+#
+# See the Licence for the specific language governing permissions and limitations
+# under the Licence.
+
+from libc.math cimport abs as cabs, floor
+
+from raysect.core cimport Node, Point2D, Point3D, rotate_z
+from raysect.optical import UnityVolumeEmitter
+from raysect.primitive import Mesh
+
+
+cdef double PI = 3.141592653589793
+
+
+cdef class Voxel(Node):
+
+    @property
+    def volume(self):
+        raise NotImplementedError()
+
+
+cdef class AxisSymmetricVoxel(Voxel):
+
+    @property
+    def cross_sectional_area(self):
+        raise NotImplementedError()
+
+
+cdef class ToroidalAnnulusVoxel(AxisSymmetricVoxel):
+
+    cdef:
+        Point2D _lower_corner, _upper_corner
+        list _voxel_primitives
+
+    def __init__(self, Point2D lower_corner, Point2D upper_corner, material=None):
+
+        if not isinstance(lower_corner, Point2D) or not isinstance(upper_corner, Point2D):
+            raise TypeError('The ToroidalAnnulusVoxel can only be specified with two Point2D objects.')
+
+        self._lower_corner = lower_corner
+        self._upper_corner = upper_corner
+
+        material = material or UnityVolumeEmitter()
+
+        radius = (upper_corner.x + lower_corner.x)/2
+        dr = upper_corner.x - lower_corner.x
+        number_segments = floor(2 * PI * radius / dr)
+        theta_adjusted = 360 / number_segments
+
+        # Set of points in x-z plane
+        p1a = Point3D(lower_corner.x, 0, lower_corner.y)  # corresponds to lower corner is x-z plane
+        p2a = Point3D(lower_corner.x, 0, upper_corner.y)
+        p3a = Point3D(upper_corner.x, 0, upper_corner.y)  # corresponds to upper corner in x-z plane
+        p4a = Point3D(upper_corner.x, 0, lower_corner.y)
+
+        # Set of points rotated away from x-z plane
+        p1b = p1a.transform(rotate_z(theta_adjusted))
+        p2b = p2a.transform(rotate_z(theta_adjusted))
+        p3b = p3a.transform(rotate_z(theta_adjusted))
+        p4b = p4a.transform(rotate_z(theta_adjusted))
+
+        vertices = [[p1a.x, p1a.y, p1a.z], [p2a.x, p2a.y, p2a.z],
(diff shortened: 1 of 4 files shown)


mattngc:
As per discussion with @jacklovell, added a self.as_sightline() convenience method to the BolometerFoil class in fb78621.

cherab/tools/observers/bolometry.py
```diff
@@ -240,23 +195,19 @@ class BolometerFoil(Node):
...
-                 parent=None, render_engine=None):
-
-        self.detector_id = detector_id
+                 parent=None, units="Power", accumulate=False):
 
         # perform validation of input parameters
 
         if not isinstance(dx, float):
             raise TypeError("dx argument for BolometerFoil must be of type float.")
         if not dx > 0:
```
jacklovell: I think this is a more important condition than dx being a float. Does passing an integer dx to TargettedPixel break things? Or will it be promoted to float automatically?

cherab/tools/observers/bolometry.py
```diff
@@ -276,229 +227,151 @@ def __init__(self, detector_id, centre_point, basis_x, dx, basis_y, dy, slit,
...
+                         parent=parent, transform=translation * rotation, name=detector_id)
+
+    def __repr__(self):
+        """Returns a string representation of this BolometerFoil object."""
+        return "<BolometerFoil - " + self.name + ">"
 
     @property
     def centre_point(self):
-        return self._centre_point
+        return Point3D(0, 0, 0).transform(self.to_root())
```
jacklovell: I like this. It really makes clear what the transforms are doing. And will be good for catching erroneous transforms (like the basis_x vs basis_y rotation bug we caught a couple of weeks ago).

cherab/tools/observers/bolometry.py
```diff
@@ -276,229 +227,151 @@ def __init__(self, detector_id, centre_point, basis_x, dx, basis_y, dy, slit,
...
+            raise ValueError("The units argument of BolometerFoil must be one of 'Power' or 'Radiance'.")
 
-    @property
-    def volume_radiance_sensitivity(self):
-        if self._volume_radiance_sensitivity is None:
-            raise ValueError("The sensitivity of this BolometerFoil has not yet been calculated.")
-        return self._volume_radiance_sensitivity
+        los_observer = SightLine(pipelines=[pipeline], pixel_samples=1, quiet=True,
+                                 parent=self, name=self.name, transform=self.transform)
+        los_observer.render_engine = self.render_engine
```
jacklovell: There is no way to specify `self.render_engine` in init. So presumably we should set BolometerFoil.render_engine before calling `as_sightline`, or manually set `los_observer.render_engine` after calling this function. This should be made clear in the documentation.

cherab/tools/observers/bolometry.py
```diff
@@ -276,229 +227,151 @@ def __init__(self, detector_id, centre_point, basis_x, dx, basis_y, dy, slit,
...
+        while True:
+
+            # Find the next intersection point of the ray with the world
+            intersection = self.root.hit(CoreRay(centre_point, self.sightline_vector))
+
+            if intersection is None:
+                raise RuntimeError("No material intersection was found for this sightline.")
+
+            elif isinstance(intersection.primitive.material, NullMaterial):
+                centre_point += self.sightline_vector * 1E-4
```
jacklovell: Nice magic number. Any reason why 1E-4? Can we use the intersection primitive's width to choose an appropriate point to start the ray again instead?

cherab/tools/observers/bolometry.py
```diff
@@ -276,229 +227,151 @@ def __init__(self, detector_id, centre_point, basis_x, dx, basis_y, dy, slit,
...
+        self.min_wavelength = cached_min_wavelength
+        self.spectral_bins = cached_bins
+        self.pipelines = cached_pipelines
+        self.pixel_samples = cached_ray_count
+
+        return pipeline.samples.mean
+
+    def calculate_etendue(self, ray_count=10000, batches=10):
 
         if not batches > 5:
```
jacklovell: Why not `batches < 5` instead of `not batches > 5`? Pylint always moans at me for this. And in any case, your error message is slightly wrong. This condition as it stands is enforcing a minimum batch size of 6, not 5.

cherab/tools/observers/bolometry.py
```diff
@@ -513,181 +386,25 @@ def calculate_etendue(self, ray_count=10000, batches=10, print_results=False):
...
+                    intersection = world.hit(CoreRay(origin, direction))
 
                     if intersection is None:
                         passed += 1 * path_weight
                         break
 
                     elif isinstance(intersection.primitive.material, NullMaterial):
                         hit_point = intersection.hit_point.transform(intersection.primitive_to_world)
-                        origin = hit_point + direction * 1E-9
+                        origin = hit_point + direction * 1E-4
```
jacklovell: Again, magic number. Can this be made more robust?

cherab/tools/inversions/voxels.pyx
```diff
@@ -0,0 +1,317 @@
...
+
+
+class Voxel(Node):
+
+    @property
+    def volume(self):
+        raise NotImplementedError()
+
+
+class AxisSymmetricVoxel(Voxel):
```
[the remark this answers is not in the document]
mattngc: agreed

cherab/tools/observers/bolometry.py
mattngc: hmm... yeah I could make it test float and int together. It is important to ensure that the correct type is given, so in this case it could be an int or float. Some objects implement the > and < operators.

cherab/tools/observers/bolometry.py
mattngc: agreed

cherab/tools/observers/bolometry.py
mattngc: Yes the intention here is to bring this class more inline with the conventions of our other Observers. Raysect observers never set the render engine through the init, its always something like:
```
observer = PinholeCamera()
observer.render_engine = SerialEngine()
```

So I think this code works as expected because the SighLline observer you request will just copy all the appropriate settings from the parent BolometerFoil.

cherab/tools/observers/bolometry.py
mattngc: agreed, made the change

cherab/tools/observers/bolometry.py
mattngc: Agreed, I've refactored the code to calculate a displacement based on the minimum(dx, dy) / 100.

cherab/tools/observers/bolometry.py
mattngc: Same as above

mattngc:
We haven't forgotten about the documentation aspect of this. The priority now is to get a single version we can align all our users on. Then add documentation in rolling point by point releases.

PR #46, opened 2018-10-04
Fix cross_sectional_area calculation

jacklovell:
* Off-by-one error causing IndexError
* Need to take absolute value of sum (as specified in shoelace formula)


mattngc:
whoops...

PR #47, opened 2018-10-09
Improvements to `ToroidalVoxelGrid.plot()`

jacklovell:
- Enable plotting to an existing axes instance
- When no voxel values are provided, plot the grid cell outlines


mattngc:
Looks good.

PR #56, opened 2019-01-16
Add support for describing Axisymmetric voxels using CSG

jacklovell:
Using CSG primitives to describe Axisymmetric voxels results in
significantly lower memory usage than using meshes. The difference is
particularly pronounced for voxels with a radial width significantly
less than the mean radius of the voxel. For example, for a grid of
voxels of major radius ~1m on average:
- voxel width 5cm, CSG uses 6% the memory of mesh
- voxel width 2cm, CSG uses 2% the memory of mesh
- voxel width 2mm, 750k cell grid uses ~4.5GB of memory (mesh grid doesn't
  fit into memory on 128GB node, but a ~5000-cell subset uses ~8GB)

The reduction in memory usage makes it easier to work with high-voxel-count
grids, since they do not need to be split up. The CSG forumlation reduces the
child count of the voxels, which in turn speeds up pickling them and reduces
the file size of pickle save files.

Voxels with arbitrary polygon-shaped cross sections are supported, by splitting
the cross section into triangles. Each sub-triangle is used to form a toroidally-
symmetric sub-voxel using CSG operations on cylinders and cones, and these
sub-voxels are made children of the parent voxel.

The use of cone primitives does impact performance, as the KD tree is less
able to optimize ray tracing operations if the height of the cones are larger
than the resulting CSG voxel. As a performance enhancement, rectangular voxels
with edges on the x and z axes are given special treatment: instead of being
split into triangles, these are implemented as the subtraction of 2 cylinders.
This means that for voxel grids composed mainly of rectangular cells, with
perhaps non-rectangular or non-aligned cells at the edges of the grid, performance
is similar to the mesh formulation (but with significantly reduced memory overhead).

The choice of formulation (mesh or CSG) is chosen when the voxel is instantiated,
and defaults to mesh to maintain previous behaviour.

Also included are a series of tests. The principle of the tests is simple:
generate a series of points (r_i, 0, z_i), and check that the set of these
points which lie inside the voxel matches the set of points (r_i, z_i) which
lie in a 2D polygon of the same cross section. All possible triangle shapes
are tested, as well as a few extra cases of rectangular and pentagonal shapes.

### DIFF
--- a/cherab/tools/inversions/voxels.pyx
+++ b/cherab/tools/inversions/voxels.pyx
@@ -24,12 +24,13 @@ import matplotlib.pyplot as plt
 from matplotlib.patches import Polygon
 from matplotlib.collections import PatchCollection
 
-from raysect.core cimport Node, Point2D, Point3D, Vector3D, rotate_z, AffineMatrix3D, new_point3d
-from raysect.core.math cimport triangulate2d
+from raysect.core cimport (Node, Point2D, Vector2D, Point3D, Vector3D,
+                           rotate_z, AffineMatrix3D, new_point3d)
+from raysect.core.math cimport triangulate2d, translate, rotate_basis
 from raysect.core.math.function cimport Function3D
 from raysect.core.math.cython.utility cimport winding2d, find_index
 from raysect.core.math.random cimport uniform, point_triangle
-from raysect.primitive import Mesh
+from raysect.primitive import Mesh, Cylinder, Cone, Intersect, Subtract, Union
 from raysect.optical import UnityVolumeEmitter
 from raysect.optical cimport Spectrum, World, Primitive, Ray
 from raysect.optical.material.emitter.homogeneous cimport HomogeneousVolumeEmitter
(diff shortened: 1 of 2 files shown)


cherab/tools/inversions/voxels.pyx
```diff
@@ -49,7 +50,8 @@ class AxisymmetricVoxel(Voxel):
 
     # cdef np.ndarray _vertices, _triangles
 
-    def __init__(self, vertices, parent=None, material=None):
+    def __init__(self, vertices, parent=None, material=None,
+                 primitive_type: str = 'mesh'):
```
mattngc: We haven't been type hinting in pure python classes and code. Although this is harmless, it does break against the style conventions used everywhere else in CHERAB core. When we add a doc string we can suggest the type there instead.

mattngc:
Overall the code looks good, but I'm worried the code type hinting breaks with the style of the existing code. We could make this policy change, but it would mean updating it everywhere. Using these inconsistently would make the code messy.

Let's chat about it further and see which way we lean.

jacklovell:
I've removed the type annotations from the new methods in voxels.pyx. Should be ready to merge now.

cherab/tools/inversions/voxels.pyx
jacklovell: I've now removed the type hinting.

mattngc:
Ok, looks good to me.

PR #68, opened 2019-02-26
Allowing the ability for AtomicDataProviders to return Null rates when

mattngc:
the requested data is not available.

Linked to issue https://github.com/cherab/openadas/issues/11.


mattngc:
Decided we would implement this functionality in OpenADAS instead of in the core package. Merge no longer required.

PR #77, opened 2019-03-05
Fix reading G EQDSK files into EFITEquilibrium objects

jacklovell:
Update the import_eqdsk function to match EFITEquilibrium
instantiation. At the same time, make the function more robust
to NaN values being present in the EQDSK file, as well as the
possibility that the LCFS and/or limiter are not provided.

### DIFF
--- a/cherab/tools/equilibrium/eqdsk.py
+++ b/cherab/tools/equilibrium/eqdsk.py
@@ -18,6 +18,7 @@
 
 
 import re
+import warnings
 import numpy as np
 
 from raysect.optical import Point2D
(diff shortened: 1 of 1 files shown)


jacklovell:
For discussion: dealing with NaNs in the 1D q and f profiles is straightforward: simply drop the NaN values and rely on the interpolation provided by EFITEquilbrium when q and f profiles are requested. But for 2D profiles such as psi there are multiple options: do we either drop a whole row or column if it contains at least 1 NaN value (seems wasteful), or do we manually replace the NaN with an interpolated value before feeding it to EFITEquilibrium (more complicated, duplicate effort). Or do we just leave it, and limit support to G EQDSK files which have no NaNs in 2D data but can contain NaNs in the 1D data?

jacklovell:
Fair point. I've updated the branch to take an additional argument, and warn users that profiles with NaN values will break interpolation if they don't explicitly set `drop_nan=True`.

PR #90, opened 2019-03-29
Feature/more samplers

jacklovell:
Adds several new sampling functions. In 1D, 2D and 3D a function can now be sampled at a list of points, passed in as an N, Nx2 or Nx3 array respectively. Also, 2D and 3D functions can be sampled on a grid, where the coordinates of the rows and columns are explicitly provided, rather than derived from input ranges. The same functionality is available for functions which return a scalar or a Vector3D. Also, tests for all new functions have been added, based on the previous set of unit tests for the samplers.

### DIFF
--- a/cherab/core/math/__init__.py
+++ b/cherab/core/math/__init__.py
@@ -17,6 +17,10 @@
 # under the Licence.
 
 from .samplers import sample1d, sample2d, sample3d, samplevector2d, samplevector3d
+from .samplers import sample1d_points, sample2d_points, sample3d_points
+from .samplers import sample2d_grid, sample3d_grid
+from .samplers import samplevector2d_points, samplevector3d_points
+from .samplers import samplevector2d_grid, samplevector3d_grid
 from .function import Function1D, Function2D, Function3D, VectorFunction2D, VectorFunction3D
 from .interpolators import Interpolate1DLinear, Interpolate1DCubic
 from .interpolators import Interpolate2DLinear, Interpolate2DCubic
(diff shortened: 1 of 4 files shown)


jacklovell:
A point for discussion: at the moment the `samplexd_points` functions all expect what is essentially a 1D sequence of points, and return a 1D array of samples. I'd initially done this for simplicity, and so the output shape was predictable, but it would be pretty simple to also provide broadcasting samplers. This would enable e.g. `sample2d_broadcast(func, x, y)` where the output shape would be the result of broadcasting the `x` and `y` arrays. I'd probably just copy the relevant code from `scipy.interpolate.fitpack2._BivariateSplineBase.__call__` for `grid=False`. Do you think these extra broadcasting functions are worth adding?

Also, we now have a bit of a zoo of samplers. Can you think of a better way of naming or organising them, or is it OK to just keep them all in `samplers.pyx` with their current names?

mattngc:
Thanks for the great contribution @jacklovell :)

As for the other points, I guess lets start using them and see what kind of pattern emerges. I can see a lot of appeal to having the broadcasting samples, they would be really neat. I'm not too worried about having the extra functions, it is a utility library afterall. Nothing depends on them. It works best if its just a zoo of small but targeted functions with a clearly defined use. Even if there is a lot of them, it then allows users to quickly identify the function most relevant to them and implement it. If it was a more core function, then I definitely wouldn't want to have so many functions.

We might consider relabeling the basic sampler though to distinguish it more from the new ones.

Issue #61, opened 2019-02-14
Convert Beam bounding geometry to a Cone


mattngc:
Beam currently uses a cylinder for its bounding geometry. This leads to a lot of wasted integration. We can substantially speed up the code by switching to a Cone representation of the bounding geometry.

PR #108, opened 2019-05-13
Feature/conic beam

Mateasek:
To address the issue #61 

I have tried to replace Cylindrical bounding box of the beams by Conical bounding box. The cone axis is aligned with the beam z-axis. The cone radius was left the same as the cylinder radius (calculated from beam length and larger of the two divergences). The z-position of the apex is calculated using larger of the two divergences and 5 sigma in the focal point. Cone is rotated around x-axis by 180 degrees and shifted along z-axis to position the apex and to match the focus and beam end.


mattngc:
Thanks for this contribution @Mateasek. Assigning this one to @CnlPepper to code review and merge.

CnlPepper:
Will merge to a local feature branch for now, so we can make any minor changes if necessary before merging into development.

PR #127, opened 2019-06-25
Feature/ionisation balance

Mateasek:
I think this is now slowly getting into shape. Now I am trying to add unit tests to test all the future functionlity to allow some code diagnostics in future.


Mateasek:
I think this part is ready to be inspected.

mattngc:
Really impressed with the unit tests by the way. Nice work :)

Issue #125, opened 2019-06-24
EFITEquilibrium not properly dealing with missing limiter polygon


jacklovell:
When passing `limiter_polygon=None` into `EFITEquilibrium`, I would expect the generation of the `inside_limiter` mask to be skipped (as suggested by `_process_polygons`). However, in `__init__` the polygon is converted to a numpy float64 array, which in turn converts `None` to `nan`. This then causes the `is None` check in `_process_polygons` to fail, and so Cherab tries to build a limiter mask for a limiter of value `nan`.

We should either:

1. Only convert the limiter polygon to an array if it is not `None`.
2. Check in `_process_polygons` if the limiter polygon is `nan` rather than `None`.

I personally prefer option 1, as leaving undefined options as `None` and using the `is None` idiom seems more Pythonic. It would however mean treating the limiter differently to the other sequence-like inputs when converting them to arrays in `__init__`. Though considering the limiter is the only one which is allowed to be `None` already, perhaps this isn't such a bad thing.

Mateasek:
I also think 1. is nicer option. It is cheaper/easier to check if it is None than array of nans (and it is more pythonic as you said) plus storing whole array of nans is quite redundant.

CnlPepper:
It's a bug, it should not be attempting to convert a None into an array. I'll fix it when I can.

mattngc:
Closed in PR #134.

sgibson-mse:
Bug still present with limiter polygon reading as nans. Adding if test before array creation fixes this bug.

PR #132, opened 2019-06-28
Bug fix/limiter polygon

mattngc:
Moved the polygon array processing into _process_polygons, fixes bug #125.

@jacklovell, can you confirm this fixes your issue?


jacklovell:
Any chance you could rebase this so it doesn't have 2 other issue fixes in it?

PR #141, opened 2019-08-16
Add calculation of cross section centroid for voxels

jacklovell:
This gives the "centre of mass" of the voxel in the R, Z plane, using
a formula by Bourke which is accurate for a voxel of arbitrary polygon
cross section shape. Unit tests for this and the area calculation have
been added, as well as some typing for small performance improvements
in the algorithms.

### DIFF
--- a/cherab/tools/inversions/voxels.pyx
+++ b/cherab/tools/inversions/voxels.pyx
@@ -85,8 +85,8 @@ cdef class AxisymmetricVoxel(Voxel):
     """
 
     cdef:
-        double[:, ::1] _vertices
-        int[:, ::1] _triangles
+        public double[:, ::1] _vertices
+        public int[:, ::1] _triangles
         object _material
 
     def __init__(self, vertices, parent=None, material=None, primitive_type='csg'):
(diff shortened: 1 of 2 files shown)


jacklovell:
The cross section centroid currently fails if the voxel area is 0 (e.g. a voxel with 3 vertices, two of which are coincident). Do we want to try to support this (in which case, what should the centroid be) or should we explicitly state that only voxels with a non-zero area area supported? There are some 0-area voxels in the unit tests, which suggests that they should be supported.

jacklovell:
So, turning `AxisymmetricVoxel` into an extension class to type its attributes exposed the fact that we're using `voxel._vertices` and `voxel._triangles` in `ToroidalVoxelCollection`. I needed to make these attributes public to avoid breaking things, but typically access to private members is considered bad practice. We already have a `vertices` property on the voxels, but this returns a list of `Point2D` rather than a numpy array and so would be less performant than using the array directly. And there is no `triangles` public acessor at all. Should we leave this as it is, or have some other way of publicly providing these attributes (even if it means renaming them so they don't look like private variables)?

jacklovell:
Docstring updated with new attribute.

PR #143, opened 2019-08-22
Bugfix/emissivities from function

### LINKED ISSUE #142: ToroidalVoxelGrid emissivities from function is only sampling from one triangle
The `ToroidalVoxelGrid.emissivities_from_function` method is supposed to perform a Monte-Carlo integration of a function over the voxel cross section by sampling the function uniformly over the area bounded by the voxel cross section. This means uniformly sampling from each of the triangles which make up the voxel cross section, with the number of samples for each triangle proportional to the triangle's area. However, the current implementation is not correctly calculating the area of each triangle: only a single triangle area is computed.

While fixing this behaviour, it would also be useful to move much of the functionality of `emissivities_from_function` into the `AxisymmetricVoxel` class, and then call this method once for each voxel in the `ToroidalVoxelCollection`.

jacklovell:
Fix a bug in the emissivities_from_function method of ToroidalVoxelGrid where only the first triangle of the cross section was being sampled from. A number of other improvements have been made:

- The calculation of the emissivity from function for each voxel has been moved into the AxisymmetricVoxel object, with the ToroidalVoxelGrid simply looping through each voxel.
- Additional typing to improve performance.
- Unit tests added for a variety of AxisymmetricVoxel shapes.
- Docstring modified to explicitly describe the required symmetry for the emission function in the AxisymmetricVoxel case.
- A few small style tidy ups

Fixes #142

### DIFF
--- a/cherab/tools/inversions/voxels.pyx
+++ b/cherab/tools/inversions/voxels.pyx
@@ -54,6 +54,9 @@ cdef class Voxel(Node):
     def volume(self):
         raise NotImplementedError()
 
+    cpdef double emissivity_from_function(self, emission_function, int grid_samples=10):
+        raise NotImplementedError()
+
 
 cdef class AxisymmetricVoxel(Voxel):
     """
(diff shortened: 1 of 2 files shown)


jacklovell:
Quick question: should `ToroidalVoxelGrid.emissivities_from_function` actually live in `VoxelCollection` instead? It looks like it's now independent of the type of voxel and the symmetry of the emission function. It's up to the user to get these right, but in principle if they create a new subclass of `VoxelCollection` with a different symmetry, and an `emission_function` with the same symmetry, this should be fine.

Issue #147, opened 2019-08-29
EFITEquilibrium returns negative psi_normalised close to magnetic axis


Mateasek:
It recently happened to me that I got a negative psi_normalised close to the magnetic axis. This is imo caused by the Interpolate2DCubic function used. This could be solved by using ClampOutput1D with bottom bound equal to 0.

Mateasek:
Another issue there might be is the way the psi_normalised interpolator is generated:
``` python

self.psi_normalised = Interpolate2DCubic(r, z, (psi - psi_axis) / (psi_lcfs - psi_axis))
```
if there is not the coordinate for magnetic axis in the rz mesh, then the interpolator will not get 0 psin, but some finite positive number which will act as an incorrect interpolation limit, if then by chance you hit closer to the magnetic axis, while mapping profiles, you get an interpolation error.

This could be solved by allowing the user to switch the extrapolation to True and choose the type. This could solve both issues. Any other ideas?

Mateasek:
I absolutely confused myself... :-D

Mateasek:
I am reopening this issue because I've encountered the problem again. I loaded an equilibrium generated by fiesta code from eqdsk file. Then I sampled the region close to the magnetic axis and several obtained values of psi normalised were indeed negative. This I think should nod be allowed as it is against definition of the psi normalised. I think the 2d interpolator output should be clamped:

```python
self.psi_normalised = ClampOutput2D(Interpolate2DCubic(r, z, (psi - psi_axis) / (psi_lcfs - psi_axis)), min=0)
```

I believe this is connected to the interpolator oscillations I've encountered previously.

PR #152, opened 2019-09-16
Fix/positive psin

Mateasek:
ClampedOutput2D used to prevent psi_normalised returned by the EFITEquilibrium class to be negative. Connected to #147


mattngc:
This sounds ok to me, I've also encountered these sorts of problems before and wondered if it should be clamped. I would like @CnlPepper and @jacklovell to give reviews on this for impact before we merge.

Mateasek:
It can cause interpolation functionality provided by EFITEquilibrium to crash during

Issue #67, opened 2019-02-26
Should the vertex argument in AxisymmetricVoxel be a Nx2 array?


mattngc:
Currently the vertices argument is a list/tuple of Point2Ds. Elsewhere in the framework it is common to see Nx2 arrays as arguments, so maybe we should refactor this for consistency.

jacklovell:
This seems sensible to me. Can you make this change in such a way that it's still possible to provide a sequence of Point2Ds too, to avoid breaking existing code?

mattngc:
Yes, this should be doable. As long as its validated and handled appropriately it should be fine.

jacklovell:
Fixed by #157

PR #157, opened 2019-10-04
Enhancement/voxel performance

jacklovell:
This PR adds a few performance improvements to the creation and use of the voxels. I've typed variables where practicable, and replaced Numpy functions with cython equivalents from raysect (or hand coded) where statements are called many times with small inputs. The functionality of the code and the interfaces all remain unchanged.

With these improvements I'm seeing a modest (~40-50%) reduction in the time taken to instantiate a ToroidalVoxelCollection with a large number of voxels (~750k). Further improvements could be had if there were cythonised routines for creating primitives rather than having to use the Python constructors, but these are not currently available in raysect.


jacklovell:
Looks like the build is failing because this PR depends on the raysect development branch, rather than the latest release.

mattngc:
Hey @CnlPepper, how about we do another Raysect point release so we can unblock this? Otherwise this contrib feature will be held up for a while.

jacklovell:
I've added a new commit here which allows the `AxisymmetricVoxel` to accept an Nx2 array for its `vertices` argument. This fixes #67. I've deliberately included it as an additional commit in this PR rather than a separate PR as I think it would lead to awkward merge conflicts if based on the current development branch.

PR #165, opened 2019-10-24
Adding a basic import for FIESTA .mat equilibria files (fixes #164).

### LINKED ISSUE #164: Add Fiesta equilibrii as a supported Equilibrium type
Some parties are interested in using Fiesta equilibria in CHERAB. In theory Fiesta has very descriptive data describing the modeled equilibrium. But at the moment the only way to use a Fiesta equilibrium would be to dump to an eqdsk file and then load again. It seems more natural to add an import function for Fiesta directly.

mattngc:
Adding a FIESTA equilibrium importer.

I've tested its working for TE FIESTA files by importing and then using the built in plot_equilibrium() function. @Mateasek, can you try this with a COMPASS file?

### DIFF
--- a/cherab/tools/equilibrium/__init__.py
+++ b/cherab/tools/equilibrium/__init__.py
@@ -20,5 +20,6 @@
 from .efit import EFITEquilibrium
 from .plot import plot_equilibrium
 from .eqdsk import import_eqdsk
+from .fiesta import import_fiesta
 from .example import example_equilibrium
 # from .equ import import_equ_psi
--- a/cherab/tools/equilibrium/fiesta.py
+++ b/cherab/tools/equilibrium/fiesta.py
@@ -1,0 +1,79 @@
+
+# Copyright 2016-2018 Euratom
+# Copyright 2016-2018 United Kingdom Atomic Energy Authority
+# Copyright 2016-2018 Centro de Investigaciones Energéticas, Medioambientales y Tecnológicas
+#
+# Licensed under the EUPL, Version 1.1 or – as soon they will be approved by the
+# European Commission - subsequent versions of the EUPL (the "Licence");
+# You may not use this work except in compliance with the Licence.
+# You may obtain a copy of the Licence at:
+#
+# https://joinup.ec.europa.eu/software/page/eupl5
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the Licence is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR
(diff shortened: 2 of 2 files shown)


Mateasek:
I tried to test the function on a mat file saved from matlab by saving the structure with fiesta simulation output. The function raise an error because the file structure is different.

I asked a colleague of mine who is working with fiesta here in the institute for some information. Apparently we are using latest fiesta version (or at least updated lately) which we get from CCFE  using mercurial. He was also surprised that Fiesta has some standardized output. Could you please specify what fiesta you used and for what machine it is? Also, is the documentation you mention in the doc string accessible online?

Also the colleague said he has his own script for saving fiesta output, which became the standard here, into which he saves the tested output structure and also some additional information. Maybe the file you used is an output file created somewhere and being a machine specific standard?

mattngc:
Hey @Mateasek, can you confirm what version number you are using?

Mateasek:
We are using 8.10

mattngc:
Ok, I think I understand the issue. This was tested on a Tokamak Energy FIESTA file using version 8.6. I believe it was simply a state file of the full matlab variable space. It seems there isn't a unified format for FIESTA.

@Mateasek, can you have a quick look at the code and see if it looks like the variable name differences are minor? Otherwise, we might need to keep this code in the TE machine specific package.

Mateasek:
I think until fiesta has some standard output it would be best if it was kept in a machine specific packages. The matlab description of our output is:

``` matlab
equil =

 struct with fields:

                 label: 'shape'
   configuration_ident: 0.1419
               version: [1×1 struct]
                  date: '17-Sep-2019'
              userdata: []
                  grid: [1×1 struct]
        initial_plasma: [1×1 struct]
                 icoil: [1×1 struct]
                  irod: 2.2400e+07
                   Psi: [1×1 struct]
               Psi_vac: [1×1 struct]
                     I: [1×1 struct]
              jprofile: [1×1 struct]
               control: [1×1 struct]
              feedback: [1×1 struct]
          psi_boundary: -2.7496
              psi_axis: -1.1870
                inside: [4835×1 double]
           convergence: [1×1 struct]
                  Psi0: []

CnlPepper:
Github needs a facepalm emoji for times like this. I just mentioned to Matt on the phone that I think we'll have to come up with a Cherab equilibrium file format (most likely JSON based) at some point and leave the conversion to the labs.

mattngc:
Thanks both, yeah its really annoying that so many of these codes don't have better standards for file formats. I'll keep this feature in the machine spacific package then.

Issue #169, opened 2019-11-12
Arbitrary total radiated power profile


Mateasek:
This issue is inspired by [TotalRadiatedPower](https://github.com/cherab/core/blob/e6ae871761e708e031521188986baaf7e72dcd04/cherab/core/model/plasma/total_radiated_power.pyx#L25) plasma model. It calculates the quantity from ionic and electron densities and atomic data.

It would be nice to have a similar function to which you could provide a Function3D object providing the total radiated power. This is handy when you are working with outputs of transport codes, which can directly provide the profile in some dimension. This is useful when you don't know the profiles of ion densities for example.

p.s. I have a working example ready, I just need to add docstrings, pep8, comments and etc.

Mateasek:
This functionality is already available, see: `cherab.tools.emitters.RadiationFunction`

PR #170, opened 2019-11-12
Add ArbitraryTotalRadiatedPower

Mateasek:
This plasma model allows the user to specify total radiated power with a custom Function3D object which is independent of other plasma properties and atomic data. This is handy if you have an external source of the quantity for example a transport code.

Linked to #169

### DIFF
--- a/cherab/core/model/plasma/total_radiated_power.pxd
+++ b/cherab/core/model/plasma/total_radiated_power.pxd
@@ -17,6 +17,9 @@
 # under the Licence.
 
 
+from raysect.core.math.function cimport Function3D
+from raysect.optical cimport Spectrum, Point3D, Vector3D
+
 from cherab.core.atomic.elements cimport Element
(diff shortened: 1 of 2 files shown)


jacklovell:
Thanks Matej. Note that this can also be accomplished using `cherab.tools.emitters.RadiationFunction`, which has essentially the same implementation as your plasma model. Is there a reason `RadiationFunction` isn't sufficient?

Mateasek:
The reason is I haven't found it when looking for it :-D 

`RadiationFunction` provides what I was looking for, It just didn't come to my mind to check tools.....

I would only consider adding the if condition for 0 emission and returning original spectrum, to avoid adding zeros. This might be quite frequent when used with equilibrium mapping tools.

We can I think close this PR and forget it :-)

jacklovell:
OK, makes sense. I went through almost exactly the same process when I was first looking for something like `RadiationFunction`. I got lucky enough to find it before going to the trouble of making a PR!

PR #175, opened 2019-11-14
Fix Discrete2DMesh import

Mateasek:
Discrete2D moved from location "from raysect.core.math.interpolators cimport Discrete2DMesh"  to "from raysect.core.math.function cimport Discrete2DMesh"

this should make development branches of Raysect and CHERAB compatible

linked to #175

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ python:
   - "3.6"
 install:
   - pip install cython>=0.28
-  - pip install numpy scipy matplotlib raysect>=0.5.6
+  - pip install numpy scipy matplotlib raysect>=0.6.0
   - dev/build.sh
 # command to run tests
 script: dev/test.sh
--- a/cherab/core/math/mask.pxd
+++ b/cherab/core/math/mask.pxd
@@ -18,8 +18,7 @@
 # See the Licence for the specific language governing permissions and limitations
 # under the Licence.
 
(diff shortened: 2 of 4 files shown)


Mateasek:
Cheers, I've changed the import to import the Function2D from Raysect as pointed out.

jacklovell:
Thanks. I have one last request, which I've just noticed from looking at the build log: this now technically depends on `raysect>=0.6.0`, but the Travis config and pip installer data still say the project requires an older raysect version 0.5.6. Your build got lucky and got away with this because the most recent raysect was pulled in, but installation will fail for users who already have raysect 0.5.6 installed and try to install the latest cherab.

Could you please modify setup.py, requirements.txt and .travis.yml to point to the latest version of raysect? This should then be good to merge.

Mateasek:
Good spot, I changed the setup.py to require raysect >= 0.6.0

jacklovell:
Thanks. How about requirements.txt and .travis.yml? Both also still point to the older raysect version.

Mateasek:
I am not really making it straightforward, am I? Added also changes to .travis.yml and requirements.txt

jacklovell:
Perfect, thanks. @mattngc are you happy to merge this? We'll need to rebase all other open PRs onto development after this is merged, to ensure they build properly.

PR #199, opened 2020-01-07
Fix: wrong orders of constants

Mateasek:
I forgot to add orders to the constants of electron rest mass and classical radius. I am a bit embarrassed  now ...


mattngc:
Lol, whoops.

mattngc:
Scipy has updated all their constants... which has invalidated all our tests. We will update all the constants in a separate PR, see #200.

PR #202, opened 2020-01-08
WIP:Feature/thomson scattering

Mateasek:
Updated version of the TS model.


jacklovell:
400 commits? Can you please merge with the current development branch and squash to fewer commits. Looks like quite a lot of this is duplicating the project's previous history.

Mateasek:
I've noticed. I'm trying to sort it out now, but it seems I need to expand my git knowledge. Honestly I have no idea how to do this straight away...

jacklovell:
I would do:
```
git fetch origin
git merge origin/development
git reset origin/development
```
This will leave you in a state where only your changes compared with development are present. You can then do multiple `git add` and `git commit` states to commit logically separate bits of work, or just do a single commit if that's more appropriate. I think I'd prefer a single commit, if you can't produce a series of commits where the code is functional at each stage.

Mateasek:
thanks for the tips @jacklovell

PR #203, opened 2020-01-09
WIP: ThomsonScattering model

Mateasek:


cherab/core/laser/models/laserspectrum_base.pxd
```diff
@@ -0,0 +1,24 @@
...
+from raysect.core.math.function cimport Function1D
+
+from cherab.core.laser.node cimport Laser
+from cherab.core.utility.constants cimport SPEED_OF_LIGHT, PLANCK_CONSTANT
+from numpy cimport ndarray
+
+
+cdef double CONSTANT_HC = SPEED_OF_LIGHT * PLANCK_CONSTANT
+
+cdef class LaserSpectrum_base(Function1D):
```
CnlPepper: Not PEP8 compliant class name LaserSpectrumBase would be better.... but why not just LaserSpectrum?

cherab/core/laser/models/laserspectrum_base.pyx
```diff
@@ -0,0 +1,125 @@
...
+
+        self._bins = value
+        self._update_cache()
+
+    @property
+    def wavelengths(self):
+        return self._wavelengths
+
+    @property
+    def psd(self):
```
CnlPepper: PSD? not a very descriptive name.

cherab/core/laser/models/model.pyx
```diff
@@ -0,0 +1,145 @@
...
+        return self._polarization_vector
+
+    cpdef double get_beam_width2(self, double z, double wavelength):
+        return self._waist2 + wavelength ** 2 * self._waist_const * (z - self._focus_z) ** 2
+
+    @property
+    def m2(self):
+        return self._m2
+    @m2.setter
+    def m2(self, value):
```
CnlPepper: what is m2?

cherab/core/laser/scattering.pxd
```diff
@@ -0,0 +1,42 @@
...
+from raysect.optical.spectrum cimport Spectrum
+from cherab.core.laser.node cimport Laser
+from cherab.core.laser.models.model_base cimport LaserModel
+from cherab.core.laser.models.laserspectrum_base cimport LaserSpectrum_base
+from cherab.core cimport Plasma
+cimport cython
+
+cdef double RE_SQUARED
+
+cdef class ScatteringModel_base:
```
CnlPepper: Not PEP8 compliant.

mattngc:
This is new core code so both @CnlPepper and @mattngc should be involved with the review. From what I can see I recommend you start with a basic cleanup to make it PEP8 compliant. You should be able to do this easily with the warnings in pyCharm.

cherab/core/laser/models/laserspectrum_base.pyx
Mateasek: Power Spectral Density, I thought it was quite a regular abbreviation, I can change it to PowerSpectralDensity if it is more appropriate.

cherab/core/laser/models/model.pyx
Mateasek: m2 (m^2) is a factor for deviation of a gaussan laser beam from an ideal case. It should be quite known constant in laser field, or at least I hit it a lot when I did some reading about Gaussian beam description.

cherab/core/laser/models/laserspectrum_base.pxd
Mateasek: Renamed to LaserSpectrum

cherab/core/laser/scattering.pxd
Mateasek: renamed to ScatteringModel

cherab/core/laser/models/laserspectrum_base.pyx
jacklovell: I don't think PSD is a particularly uncommon term, and if it's in the class's docstring I'd be fine with it. Though if you do rename it then PEP8 would imply `power_spectral_density`.

cherab/core/laser/models/laserspectrum.pyx
```diff
@@ -0,0 +1,56 @@
...
+cdef class GaussianSpectrum(LaserSpectrum):
+
+    def __init__(self, double min_wavelength, double max_wavelength, int bins, double mu, double sigma):
+
+        self.sigma = sigma
+        self.mu = mu
+        super().__init__(min_wavelength, max_wavelength, bins)
+
+    @property
+    def sigma(self):
```
jacklovell: Can you call this `standard_deviation` instead? It would be more readable than just using the symbol typically used in written down equations involving Gaussians.

cherab/core/laser/models/laserspectrum.pyx
```diff
@@ -0,0 +1,56 @@
...
+    def sigma(self, value):
+        if not value >0:
+            raise ValueError("Value has to be larger than 0")
+
+        self._sigma = value
+        self._recip_sigma = 1 / value
+        self._normalisation = 1 / (value * sqrt(2 * M_PI))
+
+    @property
+    def mu(self):
```
jacklovell: `mean`, not `mu` please.

cherab/core/laser/models/laserspectrum_base.pyx
```diff
@@ -0,0 +1,125 @@
...
+        for index in range(self._bins):
+            self._power_mv[index] = self.evaluate(self._wavelengths_mv[index])
+            self._psd_mv[index] = self._power_mv[index] / self._delta_wavelength
+            self._photons_mv[index] = self._psd_mv[index] / self._photon_energy(self._wavelengths_mv[index])
+
+        #notify about changes
+        self._notifier.notify()
+
+    cdef double _photon_energy(self, double wavelength):
+        return CONSTANT_HC / (wavelength * 1e-9)
```
jacklovell: This is the only place you use `CONSTANT_HC`. Why not just write out `PLANCK_CONSTANT * SPEED_OF_LIGHT` in this expression instead of making yet another module-level variable? It's more readable, and will have negligible impact on performance.

cherab/core/laser/models/model.pyx
CnlPepper: Is there not a better name for it? Can you come up with one? If not this will need at least some documented.

jacklovell: I agree with @CnlPepper here. I'd rather this had a human-readable name.

cherab/core/laser/models/laserspectrum_base.pyx
CnlPepper: Unless it is particularly onerous I think it should be spelled out in full, we do that elsewhere. It helps people unfamiliar with the area learn by osmosis. Jack is correct about eh PEP8 name. If you have not read PEP8, please do: https://www.python.org/dev/peps/pep-0008/ as any commit violating pep8 won't be merged.

cherab/core/laser/models/laserspectrum.pyx
CnlPepper: We do use sigma for the neutral beam width at present. I'd prefer it is consistent with that. I do however prefer mean over mu (in your other comment).

options:

1) guassian(mean, sigma)
2) gaussian(mu, sigma)
3) gaussian(mean, stddev)

ie option 1) for now.... though with hindsight 3) may have be better. We can reevaluate this at a later date. Perhaps we can discuss argument standards in the upcoming cherab meeting.

cherab/core/laser/models/model.pyx
Mateasek: it is this: https://en.wikipedia.org/wiki/M_squared

I will rename it to beam_quality_factor, although I was told by our TS people and optician that nobody calls it like that and everybody uses m2. I will add to docstring that it is m^2

cherab/core/laser/models/laserspectrum.pyx
Mateasek: Mean I agree with. I would also pick stddev for sigma, because it is more general in describing moments of functions, as @jacklovell pointed out.

cherab/core/laser/models/laserspectrum.pyx
```diff
@@ -0,0 +1,56 @@
...
+        super().__init__(min_wavelength, max_wavelength, bins)
+
+    cdef double evaluate(self, double x) except? -1e999:
+        
+        cdef:
+            double spectrum_width
+            int index
+
+        if self._min_wavelength < x < self._max_wavelength:
+            return 1.0/ (self._max_wavelength - self._min_wavelength)
```
jacklovell: Missing space between `1.0` and `/`.

cherab/core/laser/models/laserspectrum.pyx
```diff
@@ -0,0 +1,56 @@
...
+        self.mean = mean
+        super().__init__(min_wavelength, max_wavelength, bins)
+
+    @property
+    def stddev(self):
+        return self._stddev
+
+    @stddev.setter
+    def stddev(self, value):
+        if not value >0:
```
jacklovell: Missing space between `>` and `0`.

cherab/core/laser/models/laserspectrum.pyx
```diff
@@ -0,0 +1,56 @@
...
+        self._recip_stddev = 1 / value
+        self._normalisation = 1 / (value * sqrt(2 * M_PI))
+
+    @property
+    def mean(self):
+        return self._mean
+
+    @mean.setter
+    def mean(self, double value):
+        if not value >0:
```
jacklovell: Missing space.

cherab/core/laser/models/laserspectrum_base.pyx
```diff
@@ -0,0 +1,125 @@
...
+
+    @property
+    def photons(self):
+        return self._photons
+
+    @property
+    def delta_wavelength(self):
+        return self._delta_wavelength
+
+    def _check_wavelength_validity(self, min_wavelength = None, max_wavelength = None):
```
jacklovell: PEP8 suggests no space between argument, equals sign and value in function definitions and calls (unless type hints are provided, which isn't the case here). This should be e.g. `min_wavelength=None`.

cherab/core/laser/models/model.pyx
```diff
@@ -0,0 +1,145 @@
...
+    @power.setter
+    def power(self, double value):
+        if not value >0:
+            raise ValueError("Value has to be larger than 0.")
+        
+        self._power = value
+        self.set_power_density_function(Constant3D(value))
+
+cdef class GaussianBeamAxisymmetric(LaserModel):
+    def __init__(self, Laser laser = None, Vector3D polarization = Vector3D(0, 1, 0), power=1,
```
jacklovell: No spaces around equals signs, please.

cherab/core/laser/models/model.pyx
```diff
@@ -0,0 +1,145 @@
...
+
+        return self._polarization_vector
+
+    cpdef double get_beam_width2(self, double z, double wavelength):
+        return self._waist2 + wavelength ** 2 * self._waist_const * (z - self._focus_z) ** 2
+
+    @property
+    def beam_quality_factor(self):
+        return self._beam_quality_factor
+    @beam_quality_factor.setter
```
jacklovell: Line break here.

cherab/core/laser/models/model.pyx
```diff
@@ -0,0 +1,145 @@
...
+
+    @focus_z.setter
+    def focus_z(self, double value):
+        self._focus_z = value
+
+    def _change_waist_const(self):
+        self._waist_const = (1e-9 * self._beam_quality_factor / (M_PI * self._waist_radius)) ** 2
+
+cdef class AxisymmetricGaussian(LaserModel):
+    def __init__(self, Laser laser = None, laser_sigma = 0.01, Vector3D polarization = Vector3D(0, 1, 0)):
```
jacklovell: No spaces around equals signs.

cherab/core/laser/scattering.pyx
```diff
@@ -0,0 +1,199 @@
...
+                            Spectrum spectrum):
+
+        cdef:
+            double ne, te, laser_power_density, angle_pointing, angle_polarization, laser_wavelength
+            Vector3D pointing_vector, polarization_vector
+            Spectrum laser_spectrum
+            Py_ssize_t index
+        
+        ne = self._plasma.get_electron_distribution().density(position_plasma.x, position_plasma.y, position_plasma.z)
+        te = self._plasma.get_electron_distribution().effective_temperature(position_plasma.x, position_plasma.y, position_plasma.z)
```
jacklovell: Rather than make two calls to `self._plasma.get_electron_distribution()`, can you store the distribution in a local variable once and then access that each time?

cherab/core/laser/scattering.pyx
```diff
@@ -0,0 +1,199 @@
...
+        #Polarization vector (Vector of electric field of the laser) and angle between observation and polarisation
+        #polarization_vector = self._laser_model.get_polarization(position_laser.x, position_laser.y, position_laser.z)
+        #angle_polarization = direction_observation.angle(pointing_vector)
+
+        #no scattering contribution cases
+        if ne <= 0 or te <=0:
+            #print("ne = {0} or te = {1}".format(ne, te))
+            return spectrum
+        for index in range(self._laser_bins):
+            laser_power_density = self._laser_power_mv[index] * self._laser_model.get_power_density(position_laser.x, position_laser.y, position_laser.z)
```
jacklovell: The power density is independent of `index`. Move it outside the loop.

cherab/core/laser/scattering.pyx
```diff
@@ -0,0 +1,199 @@
...
+
+        angle_polarization = 90.
+        #todo: uncomment if influence of polarization angle is verified
+        #Polarization vector (Vector of electric field of the laser) and angle between observation and polarisation
+        #polarization_vector = self._laser_model.get_polarization(position_laser.x, position_laser.y, position_laser.z)
+        #angle_polarization = direction_observation.angle(pointing_vector)
+
+        #no scattering contribution cases
+        if ne <= 0 or te <=0:
+            #print("ne = {0} or te = {1}".format(ne, te))
```
jacklovell: Remove commented out debug code.

jacklovell:
There's a bit of whitespace to tidy up too: please ensure there is no trailing whitespace at the ends of lines, and that each file ends with a single newline. Your use of whitespace in files is a bit inconsistent too: please ensure classes and top level functions have 2 blank lines around them and methods have 1 blank line around them. Otherwise use whitespace sparingly. Running `flake8` on your code will help.

cherab/core/laser/scattering.pyx
Mateasek: you mean something like:
``` python
e_distribution = self._plasma.get_electron_distribution()
ne = distribution.density(position_plasma.x, position_plasma.y, position_plasma.z)
te = distribution.effective_temperature(position_plasma.x, position_plasma.y, position_plasma.z) 
```
?

I know that look-ups of functions can take time, is this the reason?Or is it just readability?

jacklovell:
I don't use Pycharm, but in the past I've found both pylint and flake8 have complained about code which Pycharm was apparently happy with (i.e. other authors have committed it). So perhaps try using these two more fussy tools to see if they throw anything else up. Use your discretion though: don't just change code to make the linter happy if it hurts readability.

cherab/core/laser/scattering.pyx
jacklovell: Yes, that's what I had in mind. I doubt it will make a significant difference to performance in this case, assuming the function is just a simple getter. But it does help readability, as it means this line of code is only doing one thing (getting the density of the distribution) rather than multiple things (first get the electron distribution of the plasma object, then get the density from that).

mattngc:
> 
> 
> I don't use Pycharm, but in the past I've found both pylint and flake8 have complained about code which Pycharm was apparently happy with (i.e. other authors have committed it). So perhaps try using these two more fussy tools to see if they throw anything else up. Use your discretion though: don't just change code to make the linter happy if it hurts readability.

Yeah I agree with Jack. But its more about consistancy. All of the files accross the framework have a fairly uniform style reguardless of who the author actually was. This is important because it makes it easy for users to navigate the code, expecting a certain structure and layout. So as Jack says, don't go overboard with the linter. Instead ensure your new contribution is consistent with the style already used across Raysect and CHERAB.

CnlPepper:
Pycharm does highlight spacing and whitespace errors. It will underline code and lines with a brown wavy line. if you hover over the line it'll tell you why.

I think I'm going to add linting to the CI tests. But I'll need to turn off some of the defaults on line length etc... which are a little excessive. Cython code will be an issue as not all the linting tools understand cython.

Issue #171, opened 2019-11-12
Bug in BolometerFoil.trace_sightline kills performance


jacklovell:
When the ray hits a `NullMaterial` primitive, the small displacement used to re-fire the ray is applied to the start point rather than the intersection hit point. This means the origin just creeps towards the intersection, and it's only after firing many rays that the displacement is large enough to take the ray origin past the intersection. This leads to massive overhead as many unnecessary rays are fired, repeatedly hitting the primitive.

The `trace_sightline` logic should match that of `calculate_etendue`. PR incoming.

jacklovell:
Fixed by #172

PR #206, opened 2020-01-24
Feature/IRVB

jacklovell:
We have resistive foil bolometers in Cherab, but as yet there isn't an IRVB which means users need to roll their own. This PR adds a `BolometerIRVB` class, which provides a very similar API to the existing foil bolometer.

There are a few things I would like to check before merging:

- I couldn't quite get the IRVB to inherit from `BolometerFoil`, as it's inherently a 2D rather than a 0D observer. But there is a lot of code in common between the two objects. I'm worried that trying to force both into inheriting from a common subclass would make things more complicated than just having two separate-but-similar classes, but perhaps @mattngc and @CnlPepper can comment on whether or not this is a good idea.
- The foil is pixelated, as this is how IRVBs are analysed. I've chosen to index the 2D pixel array as (column, row). Is this a sensible choice, and does it seem consistent with the rest of Cherab and Raysect? The image is inverted, since this is basically a pinhole camera: I intend to leave it to downstream codes to deal with this, as they often have to rotate the IR camera image of the foil anyway and this can be folded in to that analysis.
- In the demo, I've tried to make a reasonably realistic tokamak example by building a tokamak vacuum vessel out of the subtraction of an outer cylinder and a mesh created from the limiting surface of the example equilibrium. This works well, but when I try to add a port (i.e. a cylinder subtracted from the vessel) it seems rays fired from the foil hit the cap of the cylinder, even though the subtraction should mean this cap isn't treated as a surface in the resulting CSG object. I'd like to get this fixed before mergin, so that I can move the IRVB back outside the plasma volume!

I haven't yet added the demo to the online documentation, as I wanted to check that the API is suitable and fix the CSG problem first. I'll add that before the final review and merge.


demos/observers/bolometry/irvb.py
```diff
@@ -0,0 +1,300 @@
...
+eq = example_equilibrium()
+r_inner = eq.limiter_polygon[:, 0].min()
+r_outer = eq.limiter_polygon[:, 0].max()
+z_lower = eq.limiter_polygon[:, 1].min()
+z_upper = eq.limiter_polygon[:, 1].max()
+dr = r_outer - r_inner
+dz = z_upper - z_lower
+outer_wall = Cylinder(radius=3, height=3, transform=translate(0, 0, -1.5))
+first_wall = axisymmetric_mesh_from_polygon(eq.limiter_polygon, 20)
+vessel = Subtract(outer_wall, first_wall)
```
mattngc: The lines below are failing because of the clash between inside / outside the CSG compound. Its hard to see without the Raysect visualisation tools in my private branch.

The outer wall is at a much larger radius than the inner wall. The 'irvb_port' cylinder cuts through the first wall but isn't long enough to also cut through the outer wall. So what is inside / outside?

For this example you don't actually need the outer_wall subtraction. I changed these lines to this and then it worked for me. But you might want to move the camera into the proper position and let me know if it still works.

Try changing 105-107 to simply:
`vessel = axisymmetric_mesh_from_polygon(eq.limiter_polygon, 20)`

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+       >>> centre_point = Point3D(0, 0, -0.08)
+       >>> detector = BolometerIRVB("ch#1", centre_point, basis_x, basis_y, width, pixels, slit, parent=camera)
+    """
+
+    def __init__(self, detector_id, centre_point, basis_x, basis_y, width, pixels, slit,
+                 parent=None, units="Power", accumulate=False, curvature_radius=0):
+
+        # perform validation of input parameters
+
+        if not isinstance(width, (float, int)):
```
CnlPepper: Prescribing a type is not the best way to regularise data in python. Convert the type and then validate once it is in a known type. e.g. what would happen it a user had an np.float64? They would be forced to convert to a python float despite both being conceptually valid. Do the following (or equivalent)

```
# explictly convert type to type with known properties
width = float(width)

# tests follow with known type
if BLAH......
```

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+            raise TypeError("centre_point argument for BolometerIRVB must be of type Point3D.")
+
+        if not isinstance(curvature_radius, (float, int)):
+            raise TypeError("curvature_radius argument for BolometerIRVB "
+                            "must be of type float/int.")
+        if curvature_radius < 0:
+            raise ValueError("curvature_radius argument for BolometerIRVB "
+                             "must not be negative.")
+
+        if not isinstance(basis_x, Vector3D):
```
CnlPepper: Explicitly convert to Vector3D, then test properties. The user should then be able to supply e.g. a tuple to the init: (1.0, 0.0, 0.0):

```
basis_x = Vector3D(*basis_x)
```

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+            raise TypeError("The basis vectors of BolometerIRVB must be of type Vector3D.")
+
+        basis_x = basis_x.normalise()
+        basis_y = basis_y.normalise()
+        normal_vec = basis_x.cross(basis_y)
+        self._slit = slit
+        self._curvature_radius = curvature_radius
+        self._accumulate = accumulate
+
+        # setup root bolometer foil transform
```
CnlPepper: This breaks the raysect and cherab  convention of supplying an AffineMatrix to define the space relative to the parent. Please don't do this as it forces a specific representation on the user.

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+        return Vector3D(1, 0, 0).transform(self.to_root())
+
+    @property
+    def basis_y(self):
+        return Vector3D(0, 1, 0).transform(self.to_root())
+
+    @property
+    def sightline_vectors(self):
+        return np.asarray(
+            [[pixel.centre_point.vector_to(self._slit.centre_point) for pixel in pixel_column]
```
CnlPepper: Re previous comment on using a general affine transform. The centre should ideal be at the origin of the new space i.e `Point3D(0, 0, 0)`

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+    def curvature_radius(self):
+        return self._curvature_radius
+
+    @property
+    def units(self):
+        return self._units
+
+    @units.setter
+    def units(self, units):
+        if units == "Power":
```
CnlPepper: Case insensitive. Help your users:
```
if units.lower() == "power":
    # do stuff
```

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+            pipeline.value.clear()
+
+    def as_sightlines(self):
+        """
+        Constructs a SightLine observer for each pixel in this bolometer.
+
+        :return: A 2D array of Sightline objects.
+        """
+
+        if self.units == "Power":
```
CnlPepper: This string logic should not be repeated, it's crappy to maintain. DRY principle. If you must continue this approach, use constants if these things turn up in multiple places:

```
UNITS_POWER = 'power'
UNITS_RADIANCE = 'radiance'
```

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+            pipeline = PowerPipeline2D
+        else:
+            pipeline = RadiancePipeline2D
```
CnlPepper: This means 'Power' or anything else., not ''Power' or 'Radiance'. Be explicit. Anything unexpected must be an error.

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+
+        :param rows: which pixel rows to trace. Can be an integer, a
+          list of integers or "all".
+        :param columns: which pixel columns to trace. Can be an
+          integer, a list of integers or "all".
+
+        :return: A 2D array of tuples containing the origin point, hit
+          point and terminating surface primitive for each pixel.
+        """
+        if rows == "all":
```
CnlPepper: What was wrong with None as a wildcard? :)

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+        :return: A 3D array of sensitivities (ncol, nrow, nvoxels)
+        """
+        # This method exploits ToroidalVoxelCollection.set_active("all"), which
+        # makes each voxel emit a different wavelength of light. By observing
+        # the voxel collection with a spectral pipeline we can thus distinguish
+        # the amount of emission from each individual voxel.
+        if not isinstance(voxel_collection, VoxelCollection):
+            raise TypeError("voxel_collection must be of type VoxelCollection")
+
+        if self.units == "Power":
```
CnlPepper: 3rd appearance of an messy bit of string handling. Apply DRY (Don't Repeat Yourself).

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+            pipeline = SpectralPowerPipeline2D(display_progress=False)
+        elif self.units == "Radiance":
+            pipeline = SpectralRadiancePipeline2D(display_progress=False)
+        else:
+            raise ValueError("Sensitivity units can only be of type 'Power' or 'Radiance'.")
```
CnlPepper: See previous comment.

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+
+        self.max_wavelength = cached_max_wavelength
+        self.min_wavelength = cached_min_wavelength
+        self.spectral_bins = cached_bins
+        self.pipelines = cached_pipelines
+        self.pixel_samples = cached_ray_count
+
+        return pipeline.frame.mean
+
+    def calculate_etendue(self, ray_count=10000, batches=10, max_distance=1e999):
```
CnlPepper: Defaults should not really be in the definition (I know there are places this still exists that are my fault, but be better than me). Do the following:

```
def calculate_etendue(self, ray_count=None, batches=None, max_distance=None):

    ray_count = ray_count or 10000
    batches = batches or 10
    max_distance = max_distance or 1e999
```

If this class is later used as a base class it'll allow the sub-class to pass through None and obtain the defaults.

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+            foil = pixels_flattened[pixel_index]
+            foil.render_engine.processes = 1
+            return pixel_index, foil.calculate_etendue(ray_count, batches, max_distance)
+
+        def update(result):
+            index, (etendue, etendue_error) = result
+            etendues_flattened[index] = etendue
+            etendue_errors_flattened[index] = etendue_error
+
+        self.render_engine.run(list(range(nx * ny)), calculate, update)
```
CnlPepper: Elegant use of nested functions to use the render engine. You have earned 1 free espresso from our office.

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
 
 
+class BolometerIRVB(TargettedCCDArray):
+    """
+    A rectangular infra red video bolometer (IRVB).
+
+    Can be configured to sample a single ray per pixel, or fan of rays
+    oriented along the observer's z axis.
+
+    :param str detector_id: The name for this detector.
```
CnlPepper: What was wrong with `name` as defined for a node?

CnlPepper: This breaks consistency with the rest of the underlying framework. This ends up as 'name' anyway... so just use `name`.

cherab/tools/observers/bolometry.py
jacklovell: The `self.units` setter already checks whether this is "Power" or "Radiance", and raises an error if it's anything else. Doing the same check/raise here seems like unnecessary repetition.

cherab/tools/observers/bolometry.py
jacklovell: To me, `rows=None` implies NO rows are returned. There's a second bit of logic a user must apply: "obviously that is stupid, so None must mean 'all'", so why not just state that explicitly?

jacklovell:
@CnlPepper thanks for the comments. I agree with most of what you've suggested, but I've also tried to make the IRVB share the same API as the `BolometerFoil` as much as possible. Should I then also change the `BolometerFoil` API, so that both foil and IRVB better match the rest of the framework (good), but break all the downstream packages such as cherab-jet, cherab-mastu which use the current API for the foils (bad)?

mattngc:
@CnlPepper, a lot of the inconsistencies you point out apply to a lot of the existing code in the Bolometry tools. So I think its better to split this into two phases. First is to just add the new IRVB feature. A second wider issue should be created to bring this whole set of tools into alignment with the rest of the framework. As @jacklovell points out, a few of the interface changes you suggested will ripple on into other packages. Better to do that in a separate refactor.

CnlPepper:
@mattngc @jacklovell This is why I'm extremely fussy about API design upfront. Once an API is exposed to the world, it becomes a burden.

Have a quick look through the other packages and see what the impact would be to tweaking the API. If it is minor, fix the packages. If it isn't a minor fix, conform to the existing API and raise an issue listing the changes to be made in the future. We can then address them later in a wider refactor.

cherab/tools/observers/bolometry.py
CnlPepper: It's defensive programming. If a mistake is made elsewhere in the codebase that accidentally changes this internal state, explicitly confirming that state at the point of use will help debugging. E.g. if a pickle __setstate__ is badly implemented or someone types 'power' instead of 'Power' you will get unexpected behaviour.

This is safer:
```
if self.units == UNITS_POWER:
    pipeline = PowerPipeline2D
elif self.units == UNITS_RADIANCE:
￼    pipeline = RadiancePipeline2D
else:
    raise ValueError(f'The units attribute was set to an unrecognised value: \'{self.units}\'.')
```

cherab/tools/observers/bolometry.py
CnlPepper: I can see your logic, but it does lead to unnecessary code. I feel it would be better to define in the docstring that it'll use all rows and columns unless explicit rows or columns are specified. Also mutable types should not be used as defaults. It's a bit of a style preference thing, so only a minor comment.

CnlPepper:
Btw I have a general nag. There seems to be some switching between ' and " and the string delimiter. The framework should use ' not ". I'm aware there are some instances in old code where this is not true, but for all new code please use single quotes unless there is a special reason.

cherab/tools/observers/bolometry.py
jacklovell: I don't follow this logic. Surely if this class is overridden then the subclass can define different default values if it wishes to? I can't see any cases where your version would give different behaviour if any of the arguments aren't specified in the call.

jacklovell:
I think that rather than introducing the IRVB and then changing the API at a later date, I'll modify this PR so that the IRVB API follows what @CnlPepper suggests and is therefore different from the `BolometerFoil`. We can then modify the `BolometerFoil` API at a later date, and this will be a smaller breaking change than modifying both the `BolometerFoil` and `BolometerIRVB` APIs.

cherab/tools/observers/bolometry.py
CnlPepper: If you define subclasses that clones default values from a base class, rather than passing none, you can cause yourself future work if you then want to modify the default values across all classes. It is best to pass none back to the base class so there is only a single point in the codebase to change to alter a default value. It's just better practice for longer term maintainability.

CnlPepper: It's the DRY principle again.

demos/observers/bolometry/irvb.py
jacklovell: This worked, thanks. I didn't realise `axisymmetric_mesh_from_polygon` created a hollow mesh: it's not obvious from the code that this is the case.

cherab/tools/observers/bolometry.py
jacklovell: OK, I remember now why I did this: bug #171 was causing very long run times for each pixel, so I wanted the caller to be able to restrict the number of pixels being traced to just the ones of interest. That bug was fixed by #172 and the sightline tracing is now very fast, so it's possible to just trace all the sightlines anyway and let the caller index the returned array for the lines they're actually interested in. For simplicity, and also for consistency with the `pixels_as_foils` and `as_sightlines` methods, I'll remove the `rows` and `columns` options entirely from this method.

cherab/tools/observers/bolometry.py
jacklovell: Done.

cherab/tools/observers/bolometry.py
jacklovell: Done.

cherab/tools/observers/bolometry.py
jacklovell: Done.

cherab/tools/observers/bolometry.py
jacklovell: Done, though `BolometerFoil` still uses `detector_id` for now.

cherab/tools/observers/bolometry.py
jacklovell: I've modified the units logic, using enum's now instead of string literals. It still looks roughly the same from the users' perspective, but internally to the class it's a bit more abstracted. I've tried to write it in such a way that adding or removing possible units values only requires modification to the Enum and some class-level dictionaries: is this the right way to go?

cherab/tools/observers/bolometry.py
```diff
@@ -687,6 +695,350 @@ def etendue_single_run(_):
...
+    :param Point3D centre_point: The centre point of the detector.
+    :param Vector3D basis_x: The x basis vector for the detector.
+    :param float width: The width of the detector along the x basis vector.
+    :param tuple pixels: The number of pixels to divide the foil into.
+      Pixels are square, so the height of the foil is determined by the
+      width of the foil and the number of rows and columns of pixels.
+    :param Node parent: The parent scenegraph node to which this detector belongs.
+      Typically a BolometerCamera() or an optical World() object.
+    :param bool units: The units in which to perform observations, can
+      be ['Power', 'Radiance'], defaults to 'Power'.
```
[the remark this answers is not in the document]
jacklovell: Done

cherab/tools/observers/bolometry.py
jacklovell: Done

cherab/tools/observers/bolometry.py
jacklovell: Done

cherab/tools/observers/bolometry.py
jacklovell: This will have to wait until the `BolometerFoil` api is fixed, I think. Other methods which treat the IRVB as a collection of foils, such as `trace_sightlines` and `as_sightlines`, use the `BolometerFoil` conventions as that's what the IRVB pixels are broken down into for these methods.

cherab/tools/observers/bolometry.py
jacklovell: This is now done: the IRVB constructor takes an affine transform instead of centre point and basis vectors. The docstring gives an example of how to produce this transform for a given centre and basis, for those used to the BolometerFoil way of doing things.

jacklovell:
@CnlPepper I've addressed as many of your comments as possible. Are you happy for this (the longest outstanding PR in Cherab core) to now be merged? The outstanding issues of BolometerFoil API will be resolved in a later commit.

Issue #225, opened 2020-07-30
Faulty periodicity check in CylindricalRayTransferEmitter and RayTransferCylinder


vsnever:
Checking for periodicity in phi-direction during initialisation of CylindricalRayTransferEmitter and RayTransferCylinder may falsely raise an error if the period is non-integer.

I'll make a pull request in a moment.

vsnever:
Here is the PR, #226.

jacklovell:
Fixed by #226.

PR #226, opened 2020-07-30
Fix periodicity check in ray transfer objects for cylindrical grids

vsnever:
This is a fix for #225.

### DIFF
--- a/cherab/tools/raytransfer/emitters.pyx
+++ b/cherab/tools/raytransfer/emitters.pyx
@@ -418,7 +418,7 @@ cdef class CylindricalRayTransferEmitter(RayTransferEmitter):
                  double rmin=0):
 
         cdef:
-            double def_integration_step, period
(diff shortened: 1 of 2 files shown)


vsnever:
Failed due to this:
`ERROR: No matching distribution found for raysect==0.7`

jacklovell:
Thanks for this Vlad. That error is expected: cherab development branch now depends on raysect development branch (the function framework was updated in a backwards-incompatible way) and the latter isn't yet available on PyPI. As long as the unit tests pass on a local checkout of both repos this will  be fine to merge.

jacklovell:
The check should be fixed now. Could you rebase this PR onto development? I'm happy to merge it once it passes CI.

vsnever:
> The check should be fixed now. Could you rebase this PR onto development? I'm happy to merge it once it passes CI.

Thank you. It's done.

PR #228, opened 2020-07-31
Do not merge: testing CI

jacklovell:
Test update of Travis CI to ensure it successfully pulls in released Raysect for merges to master.


jacklovell:
The Branch CI is expected to fail: this branch isn't named master so is being built against Raysect development (which isn't compatible with Cherab master). But the Pull Request CI should pass.

Issue #227, opened 2020-07-31
CI: which version of raysect should we build against?


jacklovell:
Currently the CI for Cherab development branch and PRs fails because it's trying to pull in Raysect 0.7. This isn't published on PyPI yet, but Cherab requires at least this release for compatibility with the function framework.

To get the build to pass, we could edit .travis.yml to use the Raysect development branch. But Cherab stable should always be compatible with raysect stable. Do we:

1. Manually edit the Travis file to pull released Raysect from PyPI every time we push to master, and then change it back to Raysect development from Github in the next commit on the development branch?
1. Add some logic to the Travis file to somehow detect if we're on master or development, and pull the appropriate Raysect branch from both?
1. Always require Cherab development and master to be compatible with Raysect master (and release 0.7)?
1. Some other solution?

CnlPepper:
Hi Jack,

I see you have been experimenting with the travis configuration. The approach you are taking with the if statement for master vs development is the approach I'd advocate. It also ensures that cherab auto-testing will spot when the raysect development has modified a feature. 

Best regards,
Alex

jacklovell:
OK, I'll pursue this. I haven't managed to get it working for PRs to master yet, but once I do I'll submit a PR to update the development branch.

jacklovell:
Fixed by #229

CnlPepper:
Loks like you'll need to be explict (>=) with the numpy version for the latest release. If you look at the log, the 1.16 version is being used as it is already present in the Travis venv:

![image](https://user-images.githubusercontent.com/8628770/89298739-a4003100-d65d-11ea-928c-236cc55a4db7.png)

CnlPepper:
See line 194 of the log.

CnlPepper:
(for other readers, see the merge request for context).

jacklovell:
Good spot. I thought Travis would spin up a fresh instance for each job, rather than re-using an old instance.

I imagine passing the `--upgrade` option to pip should ensure the newest version of Numpy will be installed over whatever's already there. I'll make another PR to check this.

jacklovell:
[...]

PR #229, opened 2020-07-31
Update CI to build against correct version of raysect

jacklovell:
As discussed in #227, there is a risk of CI builds failing if the Cherab and Raysect versions get out of sync. It is proposed that released versions of Cherab (i.e. all versions on the master branch) should be compatible with the latest publicly released version of Raysect at the time of the Cherab release. The development branch of Cherab should be compatible with the development branch of Raysect.

This PR modifies the Travis config file so that CI builds are done against the correct version of Raysect. This should ensure that all pull requests to master will be compatible with the released Raysect version, but pull requests to development and other branches will not fail.

Testing this was tricky, but #228 demonstrates that the PR is built against master for a merge to master, and this PR demonstrates a build against Raysect development.

Also, Raysect (and Cherab) depends on matplotlib, and the latest released matplotlib 3.3.0 has dropped support for Numpy <1.15. So the minimum version needs to be bumped to get all builds to pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ python:
   - "3.6"
   - "3.7"
 env:
-  - NUMPY_VERSION=1.14.6
+  - NUMPY_VERSION=1.15.0
   - NUMPY_VERSION=1.16.6
   - NUMPY_VERSION=1.18.1
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,12 @@ python:
   - "3.6"
   - "3.7"
 env:
-  - NUMPY_VERSION=1.15.0
-  - NUMPY_VERSION=1.16.6
-  - NUMPY_VERSION=1.18.1
+  - NUMPY="numpy==1.15.0"
+  - NUMPY="numpy==1.16.6"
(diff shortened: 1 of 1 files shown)


jacklovell:
While we're at it, is it worth updating the other numpy versions Cherab is built against? Or even just dropping the 1.18 one and just building against the latest numpy?

CnlPepper:
I would consider adding the latest version of numpy. You could modify the yaml to include the equality in the string to allow for any version greater than some reference version.

```
NUMPY='numpy==1.5'
NUMPY='numpy==1.6'
NUMPY='numpy>=1.8'
```
then modify the pip line:
```
pip install $NUMPY scipy matplotlib
```

Otherwise changes look fine. We'll have to be careful not to break the raysect development branch from now on.

jacklovell:
Is there a benefit to specifying `>=` rather than just `NUMPY=numpy`? As far as I can tell that should always pull in the latest version. Otherwise I'm happy to make that change and then merge this.

CnlPepper:
There isn't any real benefit, no. It just followed the pattern of the previous definitions. Feel free to use whichever.

PR #236, opened 2020-10-16
Replace electron temperature with species temperature in GaussianLine and MultipletLineShape to fix #235.

### LINKED ISSUE #235: Use effective temperature of target species instead of electrons in GaussianLine and MultipletLineShape
Currently, in `GaussianLine` and `MultipletLineShape` the thermal broadening is calculated using the electron temperature:
```python
te = self.plasma.get_electron_distribution().effective_temperature(point.x, point.y, point.z)
...
 # calculate the line width
sigma = thermal_broadening(self.wavelength, te, self.line.element.atomic_weight)
```
I suggest using the effective temperature of target species instead: 
```python
ts = self.target_species.distribution.effective_temperature(point.x, point.y, point.z)
...
 # calculate the line width
sigma = thermal_broadening(self.wavelength, ts, self.line.element.atomic_weight)
```
For ions, the use of the electron temperature does not cause a large error, but the effective temperature of neutral atoms can be an order of magnitude lower than the electron temperature in far scrape-off layer.

@mattngc, @jacklovell, @Mateasek, If you agree, I'll make this change.

vsnever:
This fixes #235 by replacing the electron temperature with the temperature of the target species when calculating thermal broadening of the spectral lines in `GaussianLine` and `MultipletLineShape`.

### DIFF
--- a/cherab/core/model/lineshape.pyx
+++ b/cherab/core/model/lineshape.pyx
@@ -168,11 +168,11 @@ cdef class GaussianLine(LineShapeModel):
     @cython.cdivision(True)
     cpdef Spectrum add_line(self, double radiance, Point3D point, Vector3D direction, Spectrum spectrum):
 
-        cdef double te, sigma, shifted_wavelength
+        cdef double ts, sigma, shifted_wavelength
         cdef Vector3D ion_velocity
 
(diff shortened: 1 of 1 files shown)


jacklovell (approved):
Looks good to me. Happy to merge if @Mateasek agrees.

Mateasek (approved):
Fine from my side

Issue #245, opened 2020-12-22
Add Zeeman spectral line shape models


vsnever:
Zeeman line shape models are currently missing in Cherab. I propose adding 3 models:

- Simple triplet model (Paschen-Back effect),
- Parametrised triplet model (also Paschen–Back effect but with additional broadening due to the line's fine structure as in [the paper](https://doi.org/10.1088/0741-3335/44/7/312) by A. Blom and C. Jupén),
- General multiplet model with wavelength and ratios of π- and σ-polarised components provided as functions of the magnetic field strength.

The multiplet model ideally should be used together with functions from cherab/adas implementing the interface to ADAS603 codes, however the cherab/adas module is not formally required.

I have already done everything except the interface to the ADAS603. I'll make a draft PR now.

vsnever:
The PR #246 adds the Zeeman line shape models, and the PR cherab/adas#4 adds the data structure required to calculate the Zeeman multiplets.

vsnever:
Fixed in #246.

Issue #237, opened 2020-10-19
Add tests for radiation models


Mateasek:
Radiation models should get tests making sure they provide the expected spectra and are not changed accidentally:

- [ ] GaussianLine 
- [ ] MultipletLineShape
- [ ] StarkBroadenedLine
- [ ] BeamCXLine
- [ ] BeamEmissionLine

mattngc:
This is a good idea. Will help with documentation and validation as well.

vsnever:
I am currently adding Zeeman spectral line shapes to Cherab along with unit tests. The `TestLineShapes` class will also include the tests for `GaussianLine`, `MultipletLineShape` and `StarkBroadenedLine`, because all these models can be tested with the same plasma setup.

PR #246, opened 2020-12-22
Add Zeeman spectral line shape models

vsnever:
This adds three Zeeman spectral line shape models to address #245. This also adds hash and equality support to `Element` and `Line` classes (see #255).

The Zeeman line shape models include:

1. `ZeemanTriplet` for Paschen-Back effect,
2. `ParametrisedZeemanTriplet` also for Paschen-Back effect but with additional broadening due to the line's fine structure (see [A. Blom and C. Jupén, Plasma Phys. Control. Fusion 44 (2002) 1229-1241](https://doi.org/10.1088/0741-3335/44/7/312)).
3. `ZeemanMultiplet` for general Zeeman effect (ideally used along with `zeeman_structure()` method from `cherab.adas`).

All three classes inherit from the base class `ZeemanLineShapeModel`, which adds the `polarisation` property to `LineShapeModel` class. When polarisation is implemented in Raysect, Zeeman line shape models will be rewritten to support it. For now, `polarisation` parameter allows to remove π- or σ-polarised light from the spectrum as if polariser is installed.

`ParametrisedZeemanTriplet` currently only includes the parameters for the lines from the paper by A. Blom and C. Jupén. Howerever, later, I'll fit some other spectral lines available in ADAS603 with this model and will add support for more lines.

The class `ZeemanStructure` is used to initialise `ZeemanMultiplet`. The respective virtual method `zeeman_structure` is added to `AtomicData` class. This method is implemented in `cherab.adas.ADAS` and uses ADAS603 routines.

This PR also includes the tests for the following line shape models: `GaussianLine`, `MultipletLineShape`, `StarkBroadenedLine`, `ZeemanTriplet`, `ParametrisedZeemanTriplet` and `ZeemanMultiplet`, partially addressing #237.

A demo and related documentation are also included.

I'm worried about the performance of `ZeemanStructure.evaluate()`. Every time it is called, it iterates over the Python lists of `Function1D` objects, which is quite slow due to unnecessary security checks. If you have any ideas on how to implement this better, please let me know.


jacklovell:
> I'm worried about the performance of `ZeemanStructure.evaluate()`. Every time it is called, it iterates over the Python lists of `Function1D` objects, which is quite slow due to unnecessary security checks. If you have any ideas on how to implement this better, please let me know.

Have you benchmarked this to see what fraction of the total time the function spends iterating over these lists? My understanding is that Python lists are actually pretty efficient to iterate over, even from Cython. There doesn't seem to be a better way to do this than using lists: see e.g. https://stackoverflow.com/questions/33851333/cython-how-do-you-create-an-array-of-cdef-class.

vsnever:
> Have you benchmarked this to see what fraction of the total time the function spends iterating over these lists? My understanding is that Python lists are actually pretty efficient to iterate over, even from Cython.

Yes, you are right. I have benchmarked the code, and these on-assignment type checks of list items do not affect the overall performance. I was just wondering if I can use strongly typed containers in Cython, but it looks like this is not possible without switching to C++.

Mateasek:
Isn't this the same as lists of beam models in case of beams? Those are handled by model managers which check the type before adding it to the list. You can then skip the type check in the emission calculation. We could probably have this kind of manager added to tools, so we can use in more models. Now it is used in case of plasma and beam models I think.

vsnever:
> You can then skip the type check in the emission calculation.

Cython will perform the checks anyway. It does not matter if the checks were performed on initialisation or not, for Cython the types of the list items are unknown. And the code that looks like this:
```python
for i in range(number_of_lines):
    wavelength = self._wavelengths[start + i]
```
will be translated into this:
```C++
__pyx_t_5 = __pyx_v_number_of_lines;
__pyx_t_8 = __pyx_t_5;
for (__pyx_t_9 = 0; __pyx_t_9 < __pyx_t_8; __pyx_t_9+=1) {
    __pyx_v_i = __pyx_t_9;
    if (unlikely(__pyx_v_self->_wavelengths == Py_None)) {
      PyErr_SetString(PyExc_TypeError, "'NoneType' object is not subscriptable");
      __PYX_ERR(0, 103, __pyx_L1_error)
    }
    __pyx_t_10 = (__pyx_v_start + __pyx_v_i);
    if (!(likely(((PyList_GET_ITEM(__pyx_v_self->_wavelengths, __pyx_t_10)) == Py_None) || likely(__Pyx_TypeTest(PyList_GET_ITEM(__pyx_v_self->_wavelengths, __pyx_t_10), __pyx_ptype_7raysect_4core_4math_8function_5float_10function1d_4base_Function1D))))) __PYX_ERR(0, 103, __pyx_L1_error)
    __pyx_t_6 = PyList_GET_ITEM(__pyx_v_self->_wavelengths, __pyx_t_10);
    __Pyx_INCREF(__pyx_t_6);
    __Pyx_XDECREF_SET(__pyx_v_wavelength, ((struct __pyx_obj_7raysect_4core_4math_8function_5float_10function1d_4base_Function1D *)__pyx_t_6));
    __pyx_t_6 = 0;
```
However, in this particular example, the performance does not drop.

Mateasek:
My bad, I thought that there was a way how to switch type check off. Anyway I still think we should have some kind of tool to handle such lists. I opened an issue about it.

CnlPepper:
Python lists and type checks are extremely fast and absolutely negligible compared to e.g. performing an interpolation with a Function object. They really don't need optimising. I would like to point you to the unlikely() and likely() in the generated C code. These compile to hints for the processor's branch predictor, indicating the most likely execution path. As such any test will incur no penalty in normal operation, the branch prediction failure penalty only occurs if the test fails.

Always profile code, don't guess.

cherab/core/atomic/zeeman.pyx
```diff
@@ -0,0 +1,124 @@
...
+    cdef double[:, :] evaluate(self, double b, bint polarisation):
+
+        cdef int i, start, number_of_lines
+        cdef np.npy_intp multiplet_shape[2]
+        cdef double ratio_sum
+        cdef np.ndarray multiplet
+        cdef double[:, :] multiplet_mv
+        cdef Function1D wavelength, ratio
+
+        b = max(0, b)
```
CnlPepper, line 87: Why is this done?

cherab/core/atomic/zeeman.pyx
```diff
@@ -0,0 +1,124 @@
...
+            number_of_lines = self._number_of_sigma_lines
+
+        multiplet_shape[0] = 2
+        multiplet_shape[1] = number_of_lines
+        multiplet = np.PyArray_SimpleNew(2, multiplet_shape, np.NPY_FLOAT64)
+        multiplet_mv = multiplet
+
+        ratio_sum = 0
+        for i in range(number_of_lines):
+            wavelength = self._wavelengths[start + i]
```
CnlPepper, line 103: Why would you use indexing here? Just select a pi or sigma list in the if statement above. Most of this code in unnecessary if you just dynamically choose which list you are processing:

```
if polarisation == PI_POLARISATION:
    data = self._pi_data
else:
    data = self._sigma_data
````

Also, len() is fast for lists in cython, there is no need to cache the value. You are falling for the trap of over optimising vs code clarity.

File: cherab/core/model/lineshape.pyx [hunk 3]
```diff
@@ -367,6 +425,388 @@ cdef class StarkBroadenedLine(LineShapeModel):
...
+                return add_gaussian_line(radiance, shifted_wavelength, sigma, spectrum)
+
+            return add_gaussian_line(0.5 * radiance, shifted_wavelength, sigma, spectrum)
+
+        # coefficients for intensities parallel and perpendicular to magnetic field
+        cos_sqr = (b_field.dot(direction.normalise()) / b_magn)**2
+        sin_sqr = 1. - cos_sqr
+
+        # adding pi component of the Zeeman triplet
+        if self._polarisation != SIGMA_POLARISATION:
```
CnlPepper, line 531: why is the logic inverted?

```
if self._polarisation != SIGMA_POLARISATION:
```

vs 

```
if self._polarisation == PI_POLARISATION:
```

The later would be clearer.

cherab/core/model/lineshape.pyx
```diff
@@ -367,6 +425,388 @@ cdef class StarkBroadenedLine(LineShapeModel):
...
+            component_radiance = 0.5 * sin_sqr * radiance
+            spectrum = add_gaussian_line(component_radiance, shifted_wavelength, sigma, spectrum)
+
+        # adding sigma +/- components of the Zeeman triplet
+        if self._polarisation != PI_POLARISATION:
```
CnlPepper, line 536: as above

cherab/core/model/lineshape.pyx
```diff
@@ -367,6 +425,388 @@ cdef class StarkBroadenedLine(LineShapeModel):
...
+                return add_gaussian_line(radiance, shifted_wavelength, sigma, spectrum)
+
+            return add_gaussian_line(0.5 * radiance, shifted_wavelength, sigma, spectrum)
+
+        # coefficients for intensities parallel and perpendicular to magnetic field
+        cos_sqr = (b_field.dot(direction.normalise()) / b_magn)**2
+        sin_sqr = 1. - cos_sqr
+
+        # adding pi component of the Zeeman triplet
+        if self._polarisation != SIGMA_POLARISATION:
```
CnlPepper, line 706: As above.

cherab/core/model/lineshape.pyx
```diff
@@ -367,6 +425,388 @@ cdef class StarkBroadenedLine(LineShapeModel):
...
+            component_radiance = 0.5 * sin_sqr * radiance
+            spectrum = add_gaussian_line(component_radiance, shifted_wavelength, sigma, spectrum)
+
+        # adding sigma +/- components of the Zeeman triplet
+        if self._polarisation != PI_POLARISATION:
```
CnlPepper, line 711: As above.

cherab/core/model/lineshape.pyx
```diff
@@ -367,6 +425,388 @@ cdef class StarkBroadenedLine(LineShapeModel):
...
+                return add_gaussian_line(radiance, shifted_wavelength, sigma, spectrum)
+
+            return add_gaussian_line(0.5 * radiance, shifted_wavelength, sigma, spectrum)
+
+        # coefficients for intensities parallel and perpendicular to magnetic field
+        cos_sqr = (b_field.dot(direction.normalise()) / b_magn)**2
+        sin_sqr = 1. - cos_sqr
+
+        # adding pi components of the Zeeman multiplet
+        if self._polarisation != SIGMA_POLARISATION:
```
CnlPepper, line 790: Invert logic.

cherab/core/model/lineshape.pyx
```diff
@@ -367,6 +425,388 @@ cdef class StarkBroadenedLine(LineShapeModel):
...
+            component_radiance = 0.5 * sin_sqr * radiance
+            multiplet_mv = self._zeeman_structure.evaluate(b_magn, PI_POLARISATION)
+
+            for i in range(multiplet_mv.shape[1]):
+                shifted_wavelength = doppler_shift(multiplet_mv[MULTIPLET_WAVELENGTH, i], direction, ion_velocity)
+                spectrum = add_gaussian_line(component_radiance * multiplet_mv[MULTIPLET_RATIO, i], shifted_wavelength, sigma, spectrum)
+
+        # adding sigma components of the Zeeman multiplet
+        if self._polarisation != PI_POLARISATION:
```
CnlPepper, line 799: Invert logic.

CnlPepper: Shouldn't this be `else if` for safety?

cherab/core/atomic/zeeman.pyx
```diff
@@ -0,0 +1,124 @@
...
+        # normalising ratios
+        if ratio_sum > 0:
+            for i in range(number_of_lines):
+                multiplet_mv[MULTIPLET_RATIO, i] /= ratio_sum
+
+        return multiplet_mv
+
+    def __call__(self, double b, str polarisation):
+
+        if polarisation == 'pi':
```
CnlPepper, line 118: Make case insensitive with:

```
if polarisation.lower() == 'pi':
```

CnlPepper (changes requested):

cherab/core/model/lineshape.pyx
CnlPepper: Is it due to there being a 3rd valid state? If so it would be sensible to point this out in a comment.

vsnever: Yes, this is due to a 3rd valid state, `NO_POLARISATION`, which is default. In the `NO_POLARISATION` case we see both pi and sigma components of the multiplet. I will add a comment.

cherab/core/model/lineshape.pyx
vsnever: Unlike `ZeemanStructure.evaluate()` there are three options here. We could have pi-polarised components, sigma-polarised components or both if we collect our signal without a polariser. The `ZeemanStructure.evaluate()` in turn must return pi- and sigma-polarised components separately because the ratio between pi- and sigma-polarised components is determined in the `add_line()` by the angle between the magnetic field vector and the ray.

cherab/core/model/lineshape.pyx
```diff
@@ -269,43 +290,80 @@ cdef class MultipletLineShape(LineShapeModel):
...
+    :param Species target_species: The target plasma species that is emitting.
+    :param Plasma plasma: The emitting plasma object.
+    """
 
-    # Parametrised Microfield Method Stark profile coefficients.
-    # Contains embedded atomic data in the form of fits to numerical models.
-    # Only a limited range of lines is supported.
-    # See B. Lomanowski, et al. "Inferring divertor plasma properties from hydrogen Balmer
-    # and Paschen series spectroscopy in JET-ILW." Nuclear Fusion 55.12 (2015): 123028.
     STARK_MODEL_COEFFICIENTS = {
```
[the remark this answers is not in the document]
vsnever: Done.

cherab/core/model/lineshape.pyx
```diff
@@ -367,6 +425,388 @@ cdef class StarkBroadenedLine(LineShapeModel):
...
+                        line is not in the database.
+    :param float gamma: Parameter :math:`\gamma`, defaults to None. Used only if the given spectral
+                        line is not in the database.
+    :param str polarisation: Leaves only :math:`\pi`-/:math:`\sigma`-polarised components:
+                             "pi" - leave central component,
+                             "sigma" - leave side components,
+                             "no" - all components (default).
+    """
+
+    LINE_PARAMETERS = {  # alpha, beta, gamma parameters for selected lines
```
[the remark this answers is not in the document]
vsnever: Done.

cherab/core/atomic/zeeman.pyx
```diff
@@ -0,0 +1,124 @@
...
+                                   a given magnetic field strength.
+    :param list ratios_sigma: A list of Function1D objects that provide the ratios
+                              of individual :math:`\sigma`-polarised Zeeman components for a given
+                              magnetic field strength. The sum of all ratios must be equal to 1.
+    """
+
+    def __init__(self, wavelengths_pi, ratios_pi, wavelengths_sigma, ratios_sigma):
+
+        if len(wavelengths_pi) != len(ratios_pi):
+            raise ValueError('The lengths of "wavelengths_pi" ({}) and "ratios_pi" ({}) do not match.'.format(len(wavelengths_pi),
```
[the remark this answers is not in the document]
vsnever: Done. This class is now initialised with a two lists of 2-tuples: `pi_components` and `sigma_components`.

cherab/core/atomic/zeeman.pyx
vsnever: Fixed to:
```python
if b < 0:  # safety check in case if used stand-alone
    raise ValueError('Argument "b" (magnetic field strength) must be non-negative.')
```

cherab/core/atomic/zeeman.pyx
vsnever: Done.

cherab/core/model/lineshape.pyx
vsnever: Improved the comments.

cherab/core/model/lineshape.pyx
vsnever: Improved the comments here and in other similar cases.

cherab/core/atomic/zeeman.pyx
vsnever: Done.

vsnever:
Thank you very much for the review, @CnlPepper. Hope I fixed all the issues you specified. The `ZeemanStructure` class is now much clearer.

Mateasek (commented):
I hope I didn't understand the code wrongly, but I have a question about the polarisations. There is a single class calculating both polarisations and what polarisation is calculated by the ZeemanStructure.evaluate method depends on value of the polarisation argument. In lineshape models you also check the polarisation and then call the evaluate method with the correct argument. If this is correct, wouldn't it be better to have separate ZeemanStructure classes for different polarisations? This could simplify the ZeemanStructure.evaluate methods.

cherab/core/atomic/zeeman.pyx
```diff
@@ -0,0 +1,121 @@
...
+
+cimport cython
+
+np.import_array()
+
+DEF MULTIPLET_WAVELENGTH = 0
+DEF MULTIPLET_RATIO = 1
+
+DEF PI_POLARISATION = 0
+DEF SIGMA_POLARISATION = 1
```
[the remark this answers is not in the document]
vsnever: These are macros, not constant variables. However, you are right, I can define these macros inside `ZeemanStructure` class.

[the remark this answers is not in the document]
vsnever: Actually, no. I can move these definitions inside the `ZeemanStructure` class, but these macros will be available outside the class too. So, there is no point moving them.

vsnever:
>I hope I didn't understand the code wrongly, but I have a question about the polarisations. There is a single class calculating both polarisations and what polarisation is calculated by the ZeemanStructure.evaluate method depends on value of the polarisation argument. In lineshape models you also check the polarisation and then call the evaluate method with the correct argument. If this is correct, wouldn't it be better to have separate ZeemanStructure classes for different polarisations? This could simplify the ZeemanStructure.evaluate methods.

This is an option. But the `ZeemanMultiplet` class, which is initialised with the `ZeemanStructure` instant, is intended to be used along with the `cherab.adas` package, which provides the `ZeemanStructure` instants for selected spectral lines by running the ADAS603 codes. If the two different instances of `ZeemanStructure` are created for pi and sigma components, then the ADAS603 codes must be run twice. I'll think about it.

vsnever:
To support raysect/v1, `ZeemanStructure` must handle σ+ and σ- components separately, because they have different signs in `StokesVector`. I converted this PR to draft, as this can be implemented already now.

jacklovell (approved):
This PR looks good to me. @CnlPepper @Mateasek do you have any objections to merging it?

Mateasek (approved):

PR #248, opened 2020-12-25
Fix incorrect inheritance of TotalRadiatedPower from _RadiatedPower class

### LINKED ISSUE #247: Fix incorrect inheritance of TotalRadiatedPower from _RadiatedPower class
The `TotalRadiatedPower` class in `core.atomic.rates` describes the total radiated power in equilibrium conditions, which is unresolved on ionisation state. Therefore, this class must not inherit from ionisation-resolved `_RadiatedPower` class. Otherwise, the `AtomicData.total_radiated_power()` interface is broken. This should be fixed before cherab/adas#3.

vsnever:
This fixes #247. `TotalRadiatedPower` is now a base class and no longer requires an ionisation argument to initialise.

### DIFF
--- a/cherab/core/atomic/rates.pxd
+++ b/cherab/core/atomic/rates.pxd
@@ -70,6 +70,14 @@ cdef class BeamEmissionPEC(_BeamRate):
     pass
 
 
+cdef class TotalRadiatedPower():
+
+    cdef:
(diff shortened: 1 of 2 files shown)


mattngc (approved):
Looks fine to me.

PR #253, opened 2021-01-19
Correct the units for beam population and update the docs for beam-related rate coefficients to fix #252

### LINKED ISSUE #252: OpenADAS geting a wrong beam population rates
I found a possible issue in openADAS, running the [example](https://cherab.github.io/documentation/demonstrations/atomic_data/beam_plasma_interactions.html#beam-plasma-interaction-rates) on the CHERAB website, gives me different beam population rate result than presented there. 
When I run the example code i get rate ~10^-9: 
image
In the plot from the [example](https://cherab.github.io/documentation/demonstrations/atomic_data/beam_plasma_interactions.html#beam-plasma-interaction-rates) it's ~10^-3, which makes 10^6 difference, that could be unit conversion cm^3/m^3, but I am not sure as the final values are dimensionless. 
I checked the literature, the 10^-3 values are in the correct order of magnitude.
So, it seems it worked when the example was made, but does not now, at least when I run it.

vsnever:
This is a fix for #252. This PR does the following:

1. Adds normalisation argument to the `parse_adas2x_rate()`,
2. Updates `parse_adf21()`, `parse_adf22bme()` to call `parse_adas2x_rate(file, normalisation=1E-6)` and `parse_adf22bmp()` to call `parse_adas2x_rate(file, normalisation=1)`,
3. Fixes the units in `AtomicData.beam_population_rate()`, `core.atomic.rates.BeamEmissionPEC` and `core.atomic.rates.BeamCXPEC` docstrings,
4. Updates the Beam-Plasma Interaction Rates section in Cherab documentation,
5. Moves this example from beam_plasma_interactions.rst to a separate script in demos/openadas, replaces deprecated beam_cx_rate with beam_cx_pec and fixes the units in this example.

Since this PR is affecting @Mateasek's work, I suggest merging it only after we have proof from OpenADAS support, that the change made to the beam population coefficient is correct.


CnlPepper:
Looks good to me. We can wait for @Mateasek, but I'm fairly confident this is correct as I interacted with the ADAS team when developing the original models.

cherab/openadas/parse/adf21.py
```diff
@@ -19,6 +19,8 @@
 from cherab.core.utility import RecursiveDict
 from .utility import parse_adas2x_rate
 
+CM3TOM3 = 1E-6
```
Mateasek, line 22: This conversion factor is accesible from the conversion classes:

```
from cherab.core.utility.conversion import Cm3ToM3

Cm3ToM3.conversion_factor
```

Wouldn't it be better to access it this way instead of definig it again?

cherab/openadas/parse/adf22.py
```diff
@@ -19,6 +19,8 @@
 from cherab.core.utility import RecursiveDict
 from .utility import parse_adas2x_rate
 
+CM3TOM3 = 1E-6
```
Mateasek, line 22: same as for cherab/openadas/parse/adf22.py ?

Mateasek:
@vsnever thanks for the fast fix!

After @CnlPepper provided original codes and memories I think we are good to go. If I get any reply from OpenADAS which will prove this wrong, we will do another fix, but now it seems unlikely.

Mateasek (commented):
Except the minor comment about the conversion constants I think this is ready to be accepted.

CnlPepper:
For general consistency it would be a good idea to make the change @Mateasek suggests. Once that change is made, I'll merge the pull request. Thanks everyone.

CnlPepper (approved):

cherab/openadas/parse/adf21.py
vsnever: Done

cherab/openadas/parse/adf22.py
vsnever: Done

vsnever:
Thanks for reviewing the code. I fixed the issue with conversion raised by @Mateasek.

Issue #259, opened 2021-02-03
Error when calling beam_emission_pec(...)


Daniel-ler:
Hi, I think there might be possible issue in: _core/cherab/openadas/openadas.py_  
function - `beam_emission_pec(self, beam_ion, plasma_ion, charge, transition)`   - line 218.
I don't exactly understand everything that's happening in the code, so just bear with me : 
the Function call could look like this : `beam_emission_pec(Deuterium, Deuterium, 1, (3,2))`
When I do this everything works fine, I get what I believe is correct rate....
issue begins when I try to call `beam_emission_pec(Deuterium, Carbon, 6, (3,2))` then it says something like: 
_Requested wavelength (element=C, charge=5, transition=(3, 2)) is not available._

So I checked the ADAS database and this rate should be available, also, I checked local files and it was present.  
then I found that in the  `beam_emission_pec(self, beam_ion, plasma_ion, charge, transition)` 
at line 218: ` wavelength = repository.get_wavelength(plasma_ion, charge - 1, transition)`  
there is a call for wavelength, but I think it has wrong arguments, I would imagine it should look like this instead ` : 
wavelength = repository.get_wavelength(beam_ion, 0, transition)`

-as the emission is from the beam particle not the plasma. it fails at trying to get the wrong wavelength which is not present.
-also that's why the `beam_emission_pec(Deuterium, Deuterium, 1, (3,2))` works, by chance they are the same. 

But, maybe I am totally off here, because when using _core/cherab/core/model/beam/beam_emission.pyx_
I think the class `BeamEmissionLine(BeamModel)` also does the `beam_emission_pec` calls for all the plasma species,
but there seems to be no error when running, even with carbon 6 in the plasma. 

I would be grateful for any clarification on this.

Mateasek:
Hi @Daniel-ler 

thank you for reporting this. I think you isolated the bug correctly. You are correct that because it is the beam particle radiating, the get_wavelength method should get beam ion species and charge, not the plasma one. I think this smuggled itself from the CX collisions. I quickly applied the fix `repository.get_wavelength(beam_ion, 0, transition)` you suggested and it seems it is a 'hot fix' you can use in your repo.

The error does not happen only if the species are the same and the plasma species has charge 1 higher than the beam. This is why it is not happening for the case `beam_emission_pec(Deuterium, Deuterium, 1, (3,2))`

[...]

PR #262, opened 2021-02-10
Correct the get_wavelength() call in beam_emission_pec()

vsnever:
This is a hotfix for #259.

### DIFF
--- a/cherab/openadas/openadas.py
+++ b/cherab/openadas/openadas.py
@@ -215,7 +215,7 @@ class OpenADAS(AtomicData):
         try:
             # locate data file
             data = repository.get_beam_emission_rate(beam_ion, plasma_ion, charge, transition)
-            wavelength = repository.get_wavelength(plasma_ion, charge - 1, transition)
+            wavelength = repository.get_wavelength(beam_ion, 0, transition)
 
         except RuntimeError:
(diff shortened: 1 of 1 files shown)


mattngc (approved):
Thanks Vlad, looks good.

vsnever:
Looks like Travis CI is hung and will never report the status, just like in PR #246.

mattngc:
Ah, that's annoying. I was waiting for it to pass before merging. Going to bed now, but I'll try to investigate it in Travis with @CnlPepper tomorrow morning. Even though its a one line fix, better to have it pass instead of doing a forced merge.

CnlPepper:
Odd, not sure why travis has failed, I'll take a look.

CnlPepper:
So it looks like travis is no longer totally free and we have run out of our "free" credit. They have just introduced a new billing system.

CnlPepper:
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

CnlPepper:
I'll turn travis off for now, we'll need t consider an alternative or start paying.

mattngc:
Ok, thanks for investigating, sounds like the ideal time to investigate github actions as an alternative.

PR #265, opened 2021-02-11
Add more accurate wavelengths for hydrogen isotopes

vsnever:
This PR updates the wavelengths of hydrogen isotopes using the data from ADAS hdlike603 routine and NIST. Where no data was available, deuterium and tritium wavelengths were set equal to those of hydrogen and deuterium, respectively.


mattngc (approved):
Thanks for contributing Vlad, I tested a few samples and they look ok to me.

Issue #212, opened 2020-01-31
Generomak - first wall specification


mattngc:
Following issue #209, we need an importable first wall specification. The wall model should be:

- based on an octant or other repeating geometry
- saved with the package as an rsm mesh file
- a low poly first wall outline, simple low poly tiles with an example port or two

The TMC assigned @jacklovell to coordinate this task with @mattngc as reviewer.

jacklovell:
Once we have a limiter polygon from #211 I'll go ahead and make a first wall based on that.

Mateasek:
@jacklovell I talked to one of our mechanical engineers Vojtech Balner, who is now working on PFC and vessel design for COMPASS Upgrade and asked him if he was willing to make components for Generomak. He said he would be happy to contribute, he is a fan of Cherab and likes the pictures produced with it. He predicts that it should not be that much time consuming, since he spent a lot of time with this topic  already and is a CAD engineer. We would not use just scaled PFCs and vessel they do for COMPASS for copyright reasons. We would use his experience and the designs he has as a base to come up with something reasonably shaped and realistic. The point would not be to have shapes which can withstand the plasma, transients and etc., but to look realistic enough in simulations. Please let me know if you agree.

jacklovell:
This would be great, thanks.

mattngc:
> 
> 
> @jacklovell I talked to one of our mechanical engineers Vojtech Balner, who is now working on PFC and vessel design for COMPASS Upgrade and asked him if he was willing to make components for Generomak. He said he would be happy to contribute, he is a fan of Cherab and likes the pictures produced with it. He predicts that it should not be that much time consuming, since he spent a lot of time with this topic already and is a CAD engineer. We would not use just scaled PFCs and vessel they are designing for COMPASS for copyright reasons. We would use his experience and the designs he has as a base to come up with something reasonably shaped and realistic. The point would not be to have shapes which can withstand the plasma, transients and etc., but to look realistic enough in simulations. Please let me know if you agree.

This would be fantastic if you can organise this. Something light weight but realistic. You will probably need to dimension it based on the available plasma edge models since they seem to be a limiting factor.

Mateasek:
[...]

Issue #209, opened 2020-01-31
Full Generomak plasma


mattngc:
The TMC has an ambition for CHERAB to have a full range of tutorials and model demonstrations that can be run on a simple python environment without complex dependencies. The simplest way to achieve this would be to have a full plasma and Tokamak description that is self contained inside the repository. We already have an example equilibrium and bolometry observers.

Through this issue we propose the concept of Generomak. A generic hypothetical Tokamak which is completely self described within the CHERAB package. All required input data for the plasma, wall description and diagnostics is available in a simple abstracted form.

As an initial configuration set we propose:
- equilibrium (already available)
- core plasma
- edge plasma
- NBI beam
- first wall (Octant with simple low poly tiles and a few port holes)

A basic diagnostic set could include:
- bolometry (already existing)
- IRVB (in progress, #206 )
- eore spectroscopy (beam and passive)
- divertor spectroscopy

Please use this thread to propose components. The individual components will be raised and tracked in separate issues.

CnlPepper:
I'm still keep thinking @jacklovell's name of the "Mockamak" is better! :D

CnlPepper:
But the sample size of 1 layman at home suggests Generomak is easier to understand.... ok I'l stop polluting the issue queue now.

Mateasek:
As mentioned by @vsnever and @jacklovell in #268, we should decide where we will keep the data and codes connected to Generomak (out of topic, I also like Mockamak or Mockupmak more, how did I miss this?).

One proposed solution is to create cherab/generomak folder where all the importables and data would be stored. Already proposed structures [here](https://github.com/cherab/core/pull/268#issuecomment-779705860) and [here](https://github.com/cherab/core/pull/268#issuecomment-779908703).

I like the idea of having it separately, it gives a clearer way of structuring the generomak contents and expansion.

vsnever:
I would prefer `cherab.generomak` over `cherab.tools.generomak`. I think it is important that the internal structure of `cherab.generomak` can be used as an example for creating Cherab modules for real machines.

PR #268, opened 2021-02-15
Feature/generomak firstwall

Mateasek:
This PR adds first wall components to Generomak and is related to #212.

The components are saved in obj files and contain also vertex normals. I generated the meshes in Salome from 3D models made in Catia. The vertex normals were generated post-meshing in Mesh lab.

My huge thanks goes to Vojtech Balner, who created the tile models.

The first wall is created by distributing toroidally and vertically instances of the meshes contained in the obj files.


vsnever:
Thank you very much for this PR, @Mateasek.
Since we are going to add several synthetic diagnostics for the Generomak in the future, I think it will be more convenient to add it as a submodule (cherab/generomak/machine), like the Cherab modules for real devices (JET, Compass, MAST-U), than as a separate demo file.

Mateasek:
I thought that the idea was to have it in core, so you download it with demos based on Generomak.

vsnever:
> I thought that the idea was to have it in core, so you download it with demos based on Generomak.

I am not suggesting moving it from Cherab core, on the contrary I think that it should be a part of the Cherab package itself, just like the openadas module, rather than a part of the demos. Since we will have several synthetic diagnostics and a lot of demos based on these diagnostics, I think it would be more convenient to import the wall mesh and the diagnostics like:
```python
from cherab.generomak.machine import *
from cherab.generomak.cxrs import *
```

Mateasek:
Right I understand now. For some reason I thought you wanted to have it in a separate repository. I am not sure, I don't have any preference here. Moving it to separate repo would probably result in a simpler folder tree and it could also show the users how to design their own packages. On the other hand there will be also demos not relying on Generomak and then you will have two locations with demos, but that might not be an issue. I think here @mattngc, @CnlPepper, @vsnever should also express their opinion.

jacklovell:
The generomak data should live in this repository. The files themselves, and helper routines to load them, should be within the cherab package itself, perhaps under `cherab.generomak` (though could also live somewhere like `cherab.tools.generomak`, similar to the already-existing `cherab.tools.equilibrium.example_equilibrium`). We should make sure the hierarchy permits the addition of other bits of generomak such as the magnetic equilibrium, profiles and diagnostics. Perhaps the wall should live in `cherab.generomak.vessel`, then we could additionally have `cherab.generomak.equilibrium`, `cherab.generomak.profiles` and `cherab.generomak.observers`.

The helper routine `load_first_wall` should certainly be importable. Demo scripts which go in `demos` should use these routines to load the generomak machine and make plasmas, but the demos do not need to be importable themselves: they're expected to be run standalone.

I'll also add that a full discussion of the filesystem layout for these components is best done in #209.

mattngc:
Really nice work @Mateasek, I agree with the comments above. This data should live in this repository so that we can base all our demos off it. I like the suggestions from @vsnever and @jacklovell, both `cherab.generomak` and `cherab.tools.generomak` seem like sensible locations. These folders should contain the importable functions and data, while the actually generomak demo scripts should be in their own folder hierarchy under the demos folder.

Mateasek:
I agree, the listed points are very good, I applied these changes:
I created the cherab/generomak folder.
The files and folders were moved to cherab/generomak/machine

PR #270, opened 2021-02-20
EFIT vector map2d/3d methods now accept an outside LCFS argument.

### LINKED ISSUE #254: use EFITLCFSMask in FluxCoordToCartesian
[FluxCoordToCartesian.evaluate](https://github.com/cherab/core/blob/e83bb8cf1ba19d0a645b774c460347ef302e2926/cherab/tools/equilibrium/efit.pyx#L509) returns a fixed value if the coordinate is outside the LCFS. To evaluate whether the point is inside/outside a psi normalised interpolator is used and [value compared to 1](https://github.com/cherab/core/blob/e83bb8cf1ba19d0a645b774c460347ef302e2926/cherab/tools/equilibrium/efit.pyx#L518). If the coordinate is outside the psin interpolator range, errors are returned.

This could be avoided if EFITLCFSMask was used to identify the position.

CnlPepper:
Fixes #254.

### DIFF
--- a/cherab/tools/equilibrium/efit.pxd
+++ b/cherab/tools/equilibrium/efit.pxd
@@ -1,3 +1,5 @@
+# cython: language_level=3
+
 # Copyright 2016-2018 Euratom
 # Copyright 2016-2018 United Kingdom Atomic Energy Authority
 # Copyright 2016-2018 Centro de Investigaciones Energéticas, Medioambientales y Tecnológicas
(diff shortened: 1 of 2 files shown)


CnlPepper:
@Mateasek I've modified the efit class to address the difference in behaviour between the scalar and vector map functions. Could you please test them, as I no longer have access to any test data.

Mateasek:
@CnlPepper THanks for adding this. I tested it and it removed the errors in calculations which raised issue #254.

One thing, shouldn't the imports of blending functions be added to core/cherab/core/math/function/__init__.pxd and core/cherab/core/math/function/__init__.py to maintain some backwards compatibility for now, as it was done with other functions?

CnlPepper:
The blend functions used to live in the math package, the paths have already been updated.

Issue #269, opened 2021-02-19
Fix the Gaunt factor in Bremsstrahlung class


vsnever:
The bremsstrahlung radiation model implemented in Cherab uses the following empirical expression for the Gaunt factor:
https://github.com/cherab/core/blob/e83bb8cf1ba19d0a645b774c460347ef302e2926/cherab/core/model/plasma/bremsstrahlung.pyx#L92
This Gaunt factor becomes negative at Te < 1.142 eV, which can lead to negative bremsstrahlung radiation, for example, in the divertor region.

For safety reasons, this should be modified to:
```python
gaunt_factor = max(0, 0.6183 * log(te) - 0.0821)
```

However, from a physical point of view, it seems that this expression for the Gaunt factor becomes inaccurate at Te < 5.755 eV when the Gaunt factor becomes less than 1. The theory for bremsstrahlung due to collisions of non-relativistic electrons with the point ions predicts in classical approximation (low temperature limit) that the Gaunt factor is >= 1 (see e.g. J. Bekefi, "Radiation Processes in Plasmas"). Therefore, I think that a more physical correction should be:
```python
gaunt_factor = max(1, 0.6183 * log(te) - 0.0821)
```

CnlPepper:
I'm unsure of the origin of the gaunt factor equation in the code, I think Matt implemented this model. If you have something better, please feel free to rework it to be more general.

vsnever:
There is a serious bug in the calculation of bremsstrahlung radiation. In the exponent factor:
https://github.com/cherab/core/blob/e83bb8cf1ba19d0a645b774c460347ef302e2926/cherab/core/model/plasma/bremsstrahlung.pyx#L96-L97
the Planck constant must be in [eV s] instead of [J s] and the speed of light must be in [nm] instead of [m]. Therefore, instead having `exp(-1.239e4 / wvl / te)` we have `exp(-1.986e-25 / wvl / te)`, effectively throwing out the exponent from this equation.

mattngc:
Yep, it does look like an unfortunate bug. It is confusing me though because I do remember producing some demos a while ago where the scale of the Bremsstrahlung seemed sensible compared to the line radiation... Anyway its a bug and Vlad's PR fixes it.

mattngc:
Closed by PR #271

vsnever:
> It is confusing me though because I do remember producing some demos a while ago where the scale of the Bremsstrahlung seemed sensible compared to the line radiation...

[...]

PR #271, opened 2021-02-21
Fix: Wrong factors in the Bremsstrahlung radiation model (#269)

vsnever:
This PR corrects the Gaunt factor behaviour at low temperatures (it cannot drop below 1 anymore) and converts the factor in the exponent to the correct units (from [J m] to [eV nm]).


mattngc (approved):
Looks good to me.

PR #278, opened 2021-03-21
Fix incorrect effective emission rate calculation in BeamEmissionLine

### LINKED ISSUE #275: Incorrect rates used in beam_emission.pyx
In the beam_emission.pyx in the function `Spectrum emission(...)` on the line 124 there is request for the BeamEmissionPEC:
`rate = rate_func.evaluate(self._beam.get_energy(), target_density, target_temperature)`
I believe that `rate_func.evaluate()` takes the electron density as an input, however in the above code 
` target_density = distribution.density(x, y, z)` and that only returns the ion density for the given species.

So the rate call should look more like this `rate = rate_func.evaluate(self._beam.get_energy(), target_density * species.charge, target_temperature)`

Is this a correct observation ? I found it when checking my simulation results and there was discrepancy, but I am not totally sure if this is it.

vsnever:
This fixes #275 by passing the equivalent electron density and interaction energy to `BeamEmissionPEC.evaluate()` instead of the ion density and beam energy, respectively, when calculating the effective beam emission rate in `BeamEmissionLine`.

Additional fixes include:

- replacing `species.element.atomic_number` with `species.charge` in `SingleRayAttenuator`, `BeamEmissionLine` and `BeamCXLine` (in case if non-zero rates are provided for partially ionised species);
- removing unused unit conversion functions in `charge_exchange.pyx `;
- updating `beam_emission_spectrum.py` demo and related figures in the docs to reflect the changes made by the fix #236 (notice much narrower passive H-alpha line compared to the previous calculations when the T<sub>e</sub> was used instead of neutral atom temperature).


CnlPepper:
Great, thanks!

Issue #286, opened 2021-04-15
Beam.density segfaults when called without Beam.attenuator


Mateasek:
Segmentation fault is caused when the Beam.density method is called without setting the Beam.attenuator. This can happen when there are no emission models assigned, because all the checks happen when models are added. There are two options to fix this I think:

1) Add a check into the density method if the attenuator is None and throw an error. The negative side of this solution is that the check is not needed when called by the emission models, which is the majority of calls.

2) Set up a "ZeroAttenuator" as a default value which returns zero density. The plus of this is that the density method stays untouched and errors could still be raised when an emission model is being set before the attenuator.

I personally prefer the 2nd option.

vsnever:
Hi @Mateasek,

I think that setting up a `SingleRayAttenuator` as a default value is better. It looks safe despite the fact that the attenuator setter requires setting the plasma and atomic data before the attenuator. All public methods of `SingleRayAttenuator` populate the stopping cache first. If the plasma or atomic data is missing, the error will be raised already at this stage. Both the plasma and atomic data setters update the attenuator, so it's seem ok to set them after the attenuator is already set up. 

In a situation where the user has forgotten to set up the attenuator, a `ZeroAttenuator` will return zero emission or density, the cause of which may be difficult to determine. In the same situation, a `SingleRayAttenuator` will return correct values.

However, I haven't tested this and may be wrong, so please check.

Mateasek:
Hi @vsnever,

I also thought about this solution, but the error output could be misleading if the SingleRayAttenuator is set, couldn't it? Also I wasn't really comfortable with assigning a particular model as a default one. What about if the density method of the ZeroAttenuator raised an error? In that case it would be obvious where is the problem.

vsnever:
The `SingleRayAttenuator` is the only attenuator model currently implemented in Cherab, therefore it seems natural to use it as a default one. At the same time, if for example the integration step of `SingleRayAttenuator` is not optimal for a given beam and plasma, the output can be inaccurate, which can be really misleading. 

> What about if the density method of the ZeroAttenuator raised an error? In that case it would be obvious where the problem is.

Yes, I think it's better than returning zero density.

jacklovell:
Why not just make the default attenuator a `BeanAttenuator` instance? That does what you require: it raises `NotImplementedError` when the density function is called. All attenuators must be instances of this base class anyway, so rather than implement a new "do nothing" subclass you may as well use the "do nothing" base class.

Mateasek:
[...]

PR #287, opened 2021-04-16
Prevent Beam.density causing segfault

Mateasek:
This can happen when Beam.density is called before the attenuator is set. This fixes the #286


jacklovell:
Thanks @Mateasek, looks good to me.

PR #289, opened 2021-04-17
Prevent plasma, beam and atomic_data setters in BeamAttenuator and BeamModel from taking None values

vsnever:
This fixes an issue with the `plasma`, `beam` and `atomic_data` setters `BeamAttenuator` and `BeamModel` discussed in #286.


Mateasek (approved):
Hi @vsnever , thanks for the fix. 

I think this can be merged.

PR #291, opened 2021-04-25
pass data_path to repository.get* methods

### LINKED ISSUE #290: OpenADAS can't use multiple repositories
Currently it is not possible to use [OpenADAS](https://github.com/cherab/core/blob/19b5237e97231365063eef2f181ecdfbca4f8bce/cherab/openadas/openadas.py#L27) instances with other than the default repository (specified by DEFAULT_REPOSITORY_PATH). The reason is methods getting the data from the repository (e.g. [recombination_rate](https://github.com/cherab/core/blob/19b5237e97231365063eef2f181ecdfbca4f8bce/cherab/openadas/openadas.py#L80)) don't pass the self._data_path attribute as argument. I think that is a bug since the whole framework and underlying methods are ready for this treatment.

I hope nobody minds me fixing this.

Mateasek:
Fixes #290 

Could I ask @vsnever or @jacklovell to look at this PR? Thanks

### DIFF
--- a/cherab/openadas/openadas.py
+++ b/cherab/openadas/openadas.py
@@ -54,7 +54,7 @@ class OpenADAS(AtomicData):
 
         if isinstance(ion, Isotope):
             ion = ion.element
-        return repository.get_wavelength(ion, charge, transition)
+        return repository.get_wavelength(ion, charge, transition, repository_path=self._data_path)
 
     def ionisation_rate(self, ion, charge):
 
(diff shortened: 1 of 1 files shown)


vsnever (approved):
Thanks for the fix. Everything looks fine.
I'll fix the same issue in Cherab-ADAS module.

jacklovell:
Looks fine to me. Thanks.

PR #296, opened 2021-05-26
Use the isotope's wavelength repository in the OpenADAS.wavelength() method (issue #295)

### LINKED ISSUE #295: OpenADAS.wavelength() method ignores isotope's repository
The `wavelength()` method in OpenADAS class returns the wavelengths for the element ignoring the isotopes:
https://github.com/cherab/core/blob/6a7c9e581ea9dd6f60b8312a16a2324eec446a6e/cherab/openadas/openadas.py#L55-L57
Thus, this method returns identical wavelength for different hydrogen isotopes ignoring deuterium and tritium repositories.

The simplest solution would be to remove the check:
```python
    if isinstance(ion, Isotope):
        ion = ion.element
```
and access the isotope's repository. However, by default we have such repositories only for hydrogen isotopes.

I suggest to continue using the element's repository if the isotope's repository is not available:
```python
    def wavelength(self, ion, charge, transition):
        """
        :param ion: Element object defining the ion type.
        :param charge: Charge state of the ion.
        :param transition: Tuple containing (initial level, final level)
        :return: Wavelength in nanometers.
        """
        if isinstance(ion, Isotope):
            try:
                return repository.get_wavelength(ion, charge, transition, repository_path=self._data_path)
            except RuntimeError:
                print("Warning! Requested wavelength (isotope={}, charge={}, transition={}) is not available. "
                      "Will try the element's repository.".format(ion.symbol, charge, transition))
                return repository.get_wavelength(ion.element, charge, transition, repository_path=self._data_path)

        return repository.get_wavelength(ion, charge, transition, repository_path=self._data_path)
```
Is the suggested solution ok?

vsnever:
This updates `OpenADAS.wavelength()` method to use isotope's repository instead of the element's one and adds `wavelength_element_fallback` attribute to the `OpenADAS` class, which allows to use the element's wavelength when the isotope's wavelength is not available. This fixes #295.

This also moves the comments from the `OpenADAS.__init__()` to the `OpenADAS` docstring.

### DIFF
--- a/cherab/openadas/openadas.py
+++ b/cherab/openadas/openadas.py
@@ -26,19 +26,29 @@ from cherab.openadas import repository
 
 class OpenADAS(AtomicData):
     """
+    OpenADAS atomic data source.
 
+    :param str data_path: OpenADAS local repository path.
+    :param bool permit_extrapolation: If true, informs interpolation objects to allow extrapolation
+                                      beyond the limits of the tabulated data. Default is False.
+    :param bool missing_rates_return_null: If true, allows Null rate objects to be returned when
+                                           the requested atomic data is missing. Default is False.
+    :param bool wavelength_element_fallback: If true, allows to use the element's wavelength when
+                                             the isotope's wavelength is not available.
+                                             Default is False.
     """
 
-    def __init__(self, data_path=None, permit_extrapolation=False, missing_rates_return_null=False):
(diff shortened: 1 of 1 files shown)


Mateasek (approved):
Hi @vsnever, thanks for the fix. It looks good to me.

jacklovell:
Looks good to me. Thanks.

PR #300, opened 2021-06-17
Add classes for common spectroscopic instruments

### LINKED ISSUE #299: Add spectroscopic instruments
This issue is suggested by discussion in cherab/jet#21.

Classes for spectroscopic instruments such as polychromators, survey and high-resolution spectrometers, and others can simplify the setup of pipelines and spectral properties of observers. Therefore, it makes sense to add base classes for such instruments in the tools module of the core package.

vsnever:
This fixes #299. 

Classes for spectroscopic instruments such as polychromators, survey and high-resolution spectrometers, and others can simplify the setup of pipelines and spectral properties of observers.

This PR adds the following classes:

- `SpectroscopicInstrument`: base class for all instruments,
- `Polychromator`: polychromator assembly with a set of different filters,
- `PolychromatorFilter`: `InterpolatedSF` child class for symmetrical trapezoidal filter,
- `Spectrometer`: base class for spectrometers,
- `SurveySpectrometer`: survey spectrometer with a constant spectral resolution,
- `CzernyTurnerSpectrometer`: Czerny-Turner high-resolution spectrometer.


jacklovell (commented):
This looks good. Can you please add these classes to the Sphinx documentation too? It would also be good to have some unit tests and demos, but maybe these can be added once we have a generomak plasma to observe.

vsnever:
Hi @jacklovell, thanks for the review. I've added spectroscopic instruments to the Observers section of the documentation. I've also added some unit tests for `Polychromator`,` SurveySpectrometer` and `CzernyTurnerSpectrometer`. Once we have a generomak plasma, I'll add an array of lines of sight for passive spectroscopy and a demonstration similar to the KS3 demo for the JET module.

jacklovell:
Thanks @vsnever. The documentation style is a bit inconsistent with the rest of the HTML docs though. Other Cherab classes which use `@property` decorators document those properties in the class docstring using the `ivar` Sphinx directive, rather than putting docstrings on the property methods (see e.g. the `Plasma` or `EFITEquilibrium` classes). Could you please update the docstrings for these new classes to match that? I think it's important to have consistency in the HTML documentation.

vsnever:
> Thanks @vsnever. The documentation style is a bit inconsistent with the rest of the HTML docs though. Other Cherab classes which use `@property` decorators document those properties in the class docstring using the `ivar` Sphinx directive, rather than putting docstrings on the property methods (see e.g. the `Plasma` or `EFITEquilibrium` classes). Could you please update the docstrings for these new classes to match that? I think it's important to have consistency in the HTML documentation.

Not sure why I used a different style this time. I have already used the correct style before. This has now been fixed. 

By the way, if a `@property` is also an initialiser argument, we mark it using the `param`, and don't duplicate it using the `ivar`, right?

jacklovell:
> By the way, if a `@property` is also an initialiser argument, we mark it using the `param`, and don't duplicate it using the `ivar`, right?

That's correct. Generally we try to give properties the same name as the corresponding argument to the constructor. In fact, Raysect and Cherab are generally written in such a way that most of the arguments to the constructor can be supplied later as properties. See for example `Camera(parent=world)` vs `camera.parent = world`.

cherab/tools/observers/spectroscopy/instrument.py
```diff
@@ -0,0 +1,88 @@
...
+# https://joinup.ec.europa.eu/software/page/eupl5
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the Licence is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied.
+#
+# See the Licence for the specific language governing permissions and limitations
+# under the Licence.
+
+class SpectroscopicInstrument:
```
jacklovell, line 18: This base class doesn't subclass `Node` or any of the `Observer` classes, so will never be part of a scenegraph. Presumably this is because each instance will have an observer (probably fibre bundle). This is unlike every other observer in Cherab, which is treated as a scenegraph node in  order to observe anything.

I'd suggest making a note of this in the documentation, as it could cause confusion ("why doesn't `cherab.tools.observers.SpectroscopicInstrument` have a `parent` argument?"). Since its always necessary to have an observer linked to this instrument for it to be of use, why not add an `observer` property too, to save every derived class having to define their own?

jacklovell:
Strictly speaking, these spectroscopic instruments aren't actually observers (they don't live in any scene graph). So `cherab.tools.spectroscopy` is probably a better place for them than `cherab.tools.observers.spectroscopy`. The latter should be reserved for actual observers (i.e. that inherit from the rayesct observer classes).

vsnever:
> Strictly speaking, these spectroscopic instruments aren't actually observers (they don't live in any scene graph). So `cherab.tools.spectroscopy` is probably a better place for them than `cherab.tools.observers.spectroscopy`. The latter should be reserved for actual observers (i.e. that inherit from the rayesct observer classes).

I agree. I've moved spectroscopic instruments to `cherab.tools.spectroscopy`. I have also moved the documentation for the instruments to the new Spectroscopy section. The documentation now reflects that spectroscopic instruments are not connected to the scene graph and cannot observe the world. The use of spectroscopic instruments should be clearer now as I've added code samples to the docstrings.

In addition to `pipeline_properties` I've added `pipelines()` function, which returns a list of new pipelines created according to `pipeline_properties`.

cherab/tools/observers/spectroscopy/instrument.py
vsnever: > Since its always necessary to have an observer linked to this instrument for it to be of use, why not add an `observer` property too, to save every derived class having to define their own?

A single instrument can have multiple observers connected as well as a single observer can be connected to a multiple instruments. Even if we introduce the `observers` property, it's not clear to me how it can be used. For example, to setup the spectral range of the observer we must know the list of all instruments to which this observer is connected. The code sample added to the docstring for `SurveySpectrometer` now shows this case:
```python
from numpy import ceil
from raysect.optical import World
from raysect.optical.observer import FibreOptic
from cherab.tools.spectroscopy import SurveySpectrometer, Polychromator, PolychromatorFilter

world = World()
fibreoptic = FibreOptic(name="MyFibreOptic", parent=world)
# Here, the fibre optic is "connected" to both survey spectrometer and polychromator.

# setting up the polychromator
h_alpha_filter = PolychromatorFilter(656.1, name='H-alpha filter')
ciii_465nm_filter = PolychromatorFilter(464.8, name='CIII 465 nm filter')
polychromator = Polychromator([h_alpha_filter, ciii_465nm_filter], name='MyPolychromator')

# setting up the survey spectrometer
spectrometer = SurveySpectrometer(0.1, 1024, 500, name='MySpectrometer')

fibreoptic.min_wavelength = min(spectrometer.min_wavelength, polychromator.min_wavelength)
fibreoptic.max_wavelength = max(spectrometer.max_wavelength, polychromator.max_wavelength)
bin_width = min((spectrometer.max_wavelength - spectrometer.min_wavelength) / spectrometer.spectral_bins,
                (polychromator.max_wavelength - polychromator.min_wavelength) / polychromator.spectral_bins)
fibreoptic.spectral_bins = int(ceil((fibreoptic.max_wavelength - fibreoptic.min_wavelength) / bin_width))
fibreoptic.pipelines = spectrometer.pipelines() + polychromator.pipelines()
```

jacklovell:
Why is `pipelines` a function rather than a property?

cherab/tools/spectroscopy/spectrometer.py
```diff
@@ -0,0 +1,298 @@
...
+
+    :param int spectral_bins: The number of spectral samples over the wavelength range.
+    :param float reference_wavelength: Wavelength (in nm) corresponding to
+                                       the centre of reference bin.
+    :param int reference_bin: Reference bin index. Can be negative to specify the offset.
+                              Default is None (spectral_bins // 2).
+    :param str name: Spectrometer name.
+    """
+
+    def __init__(self, spectral_bins, reference_wavelength, reference_bin=None, name=''):
```
Mateasek, line 39: What is meant by the reference wavelength and/or reference bin and does it have to be specified here in the abstract class? Wouldn't it be more general to specify only min_wavelength, max_wavelength and resolution?

 Also, when having these classes, it would be great if there was an attribute for specifying a calibration function (pixel->wavelength), which might be FunctionxD, depending on the detector. I understand this would not be useful for the observations, because we can have only a constant step per bin, but it could be useful for postprocessing and comparison of measured and modelled data. It would be aiming more at people operating a diagnostic, rather than people getting already wavelength values for bins. I would definitely use such tool when working on "raw" data obtained from spectrometer,. What do you think?

Also, some imaging spectrometers can accommodate more spectra (separated in 1 or 2D on a camera chip), is this doable with the suggested architecture?

Mateasek (changes requested):

vsnever:
> Why is `pipelines` a function rather than a property?

I think this function is misnamed. The name `new_pipelines()` would do better, because that's what it actually does. This function creates new pipeline instances based on pipeline properties of the instrument. E.g., if the two observers share the same instrument, their pipelines can be initialized using the instrument:
```python
observer1.pipelines = instrument.new_pipelines()
observer2.pipelines = instrument.new_pipelines()
```
The instrument does not need to have `pipelines` property because it does not store any pipeline instances, only the observer does.

cherab/tools/spectroscopy/polychromator.py
```diff
@@ -0,0 +1,163 @@
...
+# under the Licence.
+
+import numpy as np
+from raysect.optical import InterpolatedSF
+from raysect.optical.observer import RadiancePipeline0D
+
+from .instrument import SpectroscopicInstrument
+
+
+class PolychromatorFilter(InterpolatedSF):
```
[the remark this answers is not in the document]
vsnever: Agreed. I introduced a parent `PolychromatorFilter` class that adds `min_wavelength`, `max_wavelength`, `window` and `central_wavelength` attributes to `SpectralFunction`.  `PolychromatorFilter` also ensures that its spectral function is zero outside the filter window. `TrapezoidalFilter` inherits from `PolychromatorFilter` as you suggested.

It would be nice to be able to change the spectral bins number after initialization. If we make this a property of `PolychromatorFilter`, then we must notify the `Polychromator` when it changes. In this case we have to track the polychromators to which the filter is added. I think that this would be unnecessary complication of the code, because we can already control the number of spectral bins through the `min_bins_per_window` property of the `Polychromator` object.

cherab/tools/spectroscopy/spectrometer.py
vsnever: The reference wavelength and reference pixel make sense to JET and reflect the way the spectrometer settings are stored in JPF. So, I agree that these should be removed.

Regarding the pixel->wavelength calibration, I think that the `Spectrometer` can be initialized with a sequence of calibration arrays (one array per accommodated spectrum) containing the wavelengths of the pixel borders. Namely, the values w<sup>k</sup><sub>i</sub> and w<sup>k</sup><sub>i+1</sub> define the spectral range of the pixel p<sub>i</sub> of the k-th spectrum. After the spectrum is ray-traced, we can calibrate it with: 
```python
calibrated_spectra = spectrometer.calibrate(spectrum)
```
where `calibrate(spectrum)` will do the integration:
```python
    def calibrate(self, spectrum):
        if not isinstance(spectrum, Spectrum):
            raise TypeError('Argument spectrum must be a Spectrum instance.')
        calibrated_spectra = []
        for cal_array in self.calibration_arrays:
            calibrated_spectrum = np.zeros(cal_array.size - 1)
            for i in range(cal_array.size - 1):
                calibrated_spectrum[i] = spectrum.integrate(cal_array[i], cal_array[i + 1])
            calibrated_spectra.append(calibrated_spectrum)

        return calibrated_spectra
```
With this approach we don't care if the spectrum is mapped on the chip in 1D or 2D. Of course, the machine specific spectrometers can be initialized with the known function to calculate these calibration arrays.

The `CzernyTurnerSpectrometer` will require a sequence of `(min_wavelength, num_pixels)` values instead of the calibration arrays to initialize. The calibration arrays will be calculated based on the specified parameters of the diffraction scheme.

One can control the number of bins of ray-traced spectrum through the `min_bins_per_pixel` property.

@Mateasek, what do you think of this solution?

cherab/tools/spectroscopy/polychromator.py
```diff
@@ -0,0 +1,163 @@
...
+            raise ValueError("Argument 'window' must be positive.")
+
+        flat_top = flat_top or window - 1.e-15
+
+        if flat_top <= 0:
+            raise ValueError("Argument 'flat_top' must be positive.")
+        if flat_top > window:
+            raise ValueError("Argument 'flat_top' must be less or equal than 'window'.")
+        if flat_top == window:
+            flat_top = window - 1.e-15
```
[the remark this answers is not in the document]
vsnever: I agree that properties should return the exact values specified by the user. This has now been fixed. Also the code was wrong, the correct code is `flat_top = window * (1 - 1.e-15)`. However, the case `flat_top == window` is valid because it implements a special case of an ideal rectangular filter. The `wavelengths` arrays must not contain duplicate values to avoid ambiguity at the initialization of the `InterpolatedSF` caused by the sorting. Therefore, we need to decrease the flat top interval by a value not much greater than machine epsilon.

cherab/tools/spectroscopy/spectrometer.py
vsnever: I've reworked the spectrometers according to the proposed scheme. I think they are much more useful now.

jacklovell:
> > Why is `pipelines` a function rather than a property?
> 
> I think this function is misnamed. The name `new_pipelines()` would do better, because that's what it actually does. This function creates new pipeline instances based on pipeline properties of the instrument. E.g., if the two observers share the same instrument, their pipelines can be initialized using the instrument:
> 
> ```python
> observer1.pipelines = instrument.new_pipelines()
> observer2.pipelines = instrument.new_pipelines()
> ```
> 
> The instrument does not need to have `pipelines` property because it does not store any pipeline instances, only the observer does.

Ah, I see. In that case, something like `make_pipelines` or `create_pipelines` would be suitable names for the function, as it's clear then what it does.

vsnever:
> Ah, I see. In that case, something like make_pipelines or create_pipelines would be suitable names for the function, as it's clear then what it does.

This function is now renamed to `create_pipelines()`.

vsnever:
I replaced `pipeline_properties` with `pipeline_classes` and `pipeline_kwargs` to match the `connect_pipelines()` interface of the group observers. 
Also updated `CHANGELOG.md`.

PR #302, opened 2021-06-22
Fix FluxCoordToCartesian.evaluate()

### LINKED ISSUE #301: FluxCoordToCartesian.evaluate() from efit.pyx returns wrong value
The value returned by `FluxCoordToCartesian.evaluate()` from `cherab.tools.equilibrium.efit` ignores poloidal and normal coordinates of a given vector. This happens because
```python
poloidal = new_vector3d(f.x, 0, f.z).set_length(self._poloidal.evaluate(psi))
normal = new_vector3d(-f.z, 0, f.x).set_length(self._normal.evaluate(psi))
```
just sets `poloidal` and `normal` to None. The correct code for this is:
```python
poloidal = new_vector3d(f.x, 0, f.z)
poloidal.set_length(self._poloidal.evaluate(psi))
normal = new_vector3d(-f.z, 0, f.x)
normal.set_length(self._normal.evaluate(psi))
```
The error was discovered by @mnurgaliev, I just traced it back to `FluxCoordToCartesian.evaluate()`.

I'll fix this.

vsnever:
This fixes #301 by implementing a correct use of `Vector3D.set_length()` in `FluxCoordToCartesian.evaluate()`.

### DIFF
--- a/cherab/tools/equilibrium/efit.pyx
+++ b/cherab/tools/equilibrium/efit.pyx
@@ -537,7 +537,9 @@ cdef class FluxCoordToCartesian(VectorFunction2D):
         else:
 
             toroidal = new_vector3d(0, self._toroidal.evaluate(psi), 0)
-            poloidal = new_vector3d(f.x, 0, f.z).set_length(self._poloidal.evaluate(psi))
-            normal = new_vector3d(-f.z, 0, f.x).set_length(self._normal.evaluate(psi))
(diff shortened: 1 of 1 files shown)


CnlPepper:
Whoops, that was a silly bug. Thanks for fixing it. I'm quite tempted to make the attributes of Vector3D always return self if they self modify. This would prevent these sort of mistakes and allow chaining of calls.

PR #304, opened 2021-08-15
Use new Raysect interpolators

### LINKED ISSUE #233: Severe performance drops in some scenarios if LAPACK is configured to run in multithreaded mode and MulticoreEngine is used
Calling any caching or interpolating function that uses `numpy.linalg.solve()` if `render_engine` is set to `MulticoreEngine()` may cause severe drops in performance if numpy is built with OpenBLAS and `OPENBLAS_NUM_THREADS` environment variable is not set to 1. This happens because `solve()` also runs in parallel, and when it's called, we get N parallel processes, each of which starts N parallel threads.

The solution is to set `OPENBLAS_NUM_THREADS = 1` or `OMP_NUM_THREADS = 1` before running a simulation. I think this worth mentioning in the documentation, because the user not familiar with the code might not figure it out.

Affected demos: beam_into_slab.py, beam.py, plasma-and-beam.py

### LINKED ISSUE #303: Switch to new Raysect interpolators
When new interpolators are added to Raysect (see raysect/source#393), they should replace the existing Cherab interpolators in the Cherab code. This will solve #233 and slightly improve overall performance.

Here are the classes, functions and demo scripts that need to be updated:

1. `cherab.core.model.attenuator.SingleRayAttenuator`
2. `cherab.openadas.rates.cx.BeamCXPEC`
3. `cherab.openadas.rates.beam.BeamStoppingRate`
4. `cherab.openadas.rates.beam.BeamPopulationRate`
5. `cherab.openadas.rates.beam.BeamEmissionPEC`
6. `cherab.openadas.rates.pec.ImpactExcitationPEC`
7. `cherab.openadas.rates.pec.RecombinationPEC`
8. `cherab.openadas.rates.atomic.IonisationRate`
9. `cherab.openadas.rates.atomic.RecombinationRate`
10. `cherab.openadas.rates.atomic.ThermalCXRate`
11. `cherab.openadas.rates.radiated_power.LineRadiationPower`
12. `cherab.openadas.rates.radiated_power.ContinuumPower`
13. `cherab.openadas.rates.radiated_power.CXRadiationPower`
14. `cherab.tools.equilibrium.efit.EFITEquilibrium`
15. `cherab.tools.plasmas.ionisation_balance.interpolators1d_fractional()`
16. `cherab.tools.plasmas.ionisation_balance.interpolators2d_fractional()`
17. `cherab.tools.plasmas.ionisation_balance.interpolators1d_from_elementdensity()`
18. `cherab.tools.plasmas.ionisation_balance.interpolators1d_match_plasma_neutrality()`
19. `cherab.tools.plasmas.ionisation_balance.interpolators2d_from_elementdensity()`
20. `cherab.tools.plasmas.ionisation_balance.interpolators2d_match_plasma_neutrality()`
21. `demos/equilibrium/equilibrium.py`
22. `demos/plasmas/ionisation_balance_1d.py`
23. `demos/plasmas/ionisation_balance_2d.py`

Also the mappers `IsoMapper2D` and `IsoMapper3D` refer to `Interpolate1DCubic` in their docstrings.

The new Raysect interpolators lack the following features of Cherab interpolators:

 A. Optional support for length 1 arrays.
 B. Quadratic extrapolation for cubic splines in 2D and 3D cases.
 C. Continuity of second derivative in 1D cubic spline (default option in `Interpolate1DCubic`).
Also cubic splines in Raysect and Cherab interpolate data at edge intervals differently (quadratic in Cherab vs cubic in Raysect).

Feature A is used in the items 2 - 5.
Feature B is used in the items 3 - 13.
Feature C is used in the items 2 - 5, 14, 21 - 23.

I suggest the following solution:

- Check for length 1 arrays in the items 2 - 5 explicitly and use `Constant1D` in 1D case and custom function wrappers in 2D case.
- Replace quadratic extrapolation with the nearest neighbour method or ask to add this feature in raysect/source#393 (?).
- Drop feature C. It should be used with caution anyway in case of sharp features in the data arrays.

Should we add deprecation warnings to Cherab interpolators? 

Also replacing the Cherab cubic interpolators with the Raysect ones in the item 23 causes `lsq_linear()` to freeze in `cherab.tools.plasmas.ionisation_balance_fractional_abundance_point()`. This does not happen with Cherab interpolators even if `continuity_order` is set to 1. Switching to linear interpolation for 1D data in the item 23 solves the problem.

I'll create a draft pull request with the changes I've already made to test the new Raysect interpolators.

vsnever:
This fixes #233 and #303 by replacing Cherab interpolators with the Raysect ones. The PR raysect/source#393 must be merged into development branch first.


cherab/openadas/rates/beam.pyx
```diff
@@ -19,10 +19,31 @@
...
+cdef class FunctionX2D(Function2D):
+    cdef Function1D _function
+
+    def __init__(self, Function1D function):
+        self._function = function
+
+    cdef double evaluate(self, double x, double y) except? -1e999:
+        return self._function(x)
+
+
```
jacklovell, line 46: If you are to use these, you should at least call `self._function.evaluate(x)` to avoid going through the Python `__call__` layer.

But, I think you can already accomplish this without adding to the zoo of function classes using `cherab.core.math.mappers.IsoMapper2D` and `raysect.core.math.function.float.Arg2D`:

```python
f1d_interpolator = Interpolator1DArray(...)
f2d_x_only = IsoMapper2D(Arg2D('x'), f1d_interpolator)
f2d_y_only = IsoMapper2D(Arg2D('y'), f1d_interpolator)
```

demos/plasmas/ionisation_balance_2d.py
```diff
@@ -106,13 +106,13 @@ def exp_decay(r, lamb, max_val, lim=None):
...
+n_e_1d = Interpolator1DArray(psin_1d, n_e_profile_1d, 'linear', 'none', 0)
 
-t_element_1d = Interpolate1DCubic(psin_1d, t_element_profile_1d)
-n_element_1d = Interpolate1DCubic(psin_1d, n_element_profile_1d)
-n_element2_1d = Interpolate1DCubic(psin_1d, n_element2_profile_1d)
-n_tcx_donor_1d = Interpolate1DCubic(psin_1d, n_tcx_donor_profile_1d)
+t_element_1d = Interpolator1DArray(psin_1d, t_element_profile_1d, 'linear', 'none', 0)
+n_element_1d = Interpolator1DArray(psin_1d, n_element_profile_1d, 'linear', 'none', 0)
+n_element2_1d = Interpolator1DArray(psin_1d, n_element2_profile_1d, 'linear', 'none', 0)
+n_tcx_donor_1d = Interpolator1DArray(psin_1d, n_tcx_donor_profile_1d, 'linear', 'none', 0)
```
jacklovell, line 115: Why not leave these as cubic?

cherab/openadas/rates/beam.pyx
vsnever: I didn’t think of this solution, thanks. Fixed now.

demos/plasmas/ionisation_balance_2d.py
vsnever: It appears that the bug was in `lsq_linear`, https://github.com/scipy/scipy/issues/9982. This has now been fixed, so everything works fine with the recent scipy release. I rolled back to cubic splines in this demo.

CnlPepper: I was going to say, we are not responsible for lsq_linear!

cherab/openadas/rates/atomic.pyx
```diff
@@ -33,16 +36,15 @@ cdef class IonisationRate(CoreIonisationRate):
...
 
         # store limits of data
         self.density_range = ne.min(), ne.max()
         self.temperature_range = te.min(), te.max()
 
         # interpolate rate
-        self._rate = Interpolate2DCubic(
-            ne, te, rate, extrapolate=extrapolate, extrapolation_type="quadratic"
-        )
+        extrapolation_type = 'nearest' if extrapolate else 'none'
```
CnlPepper, line 46: I'd use linear, not nearest in these classes. The rates should taper to zero. Nearest will prevent this.

vsnever: I agree that extrapolating the rates linearly makes more physical sense. Although linear extrapolation is safe to use for PEC, radiated power and beam rates, the rates of ionisation, recombination and thermal CX are interpolated and extrapolated on a log scale:
https://github.com/cherab/core/blob/ce75f778467b13ecb369b2a2bcce7ad422d4016a/cherab/openadas/rates/atomic.pyx#L36
and using linear extrapolation for exponent may be unsafe. Take for example the Be<sup>+</sup> recombination rate at Te = 1 eV. With linear extrapolation, we'll have **6.88 10<sup>-16</sup>** m<sup>3</sup> s<sup>-1</sup> for Ne = 10<sup>**22**</sup> m<sup>-3</sup> and **2.03 10<sup>6</sup>** m<sup>3</sup> s<sup>-1</sup> for Ne = 10<sup>**23**</sup> m<sup>-3</sup>.

Should these atomic rates be interpolated on a linear scale just like PEC and beam rates to avoid this problem?

jacklovell:
> I marked this as ready for review because the new interpolators are merged to Raysect development branch. However, the question remains whether Cherab interpolators should display a deprecation warning.

For the next release let's not add a deprecation warning, but instead put a note in the documentation for the Cherab interpolators to say they're deprecated and the raysect ones should be used instead. We should also update the demos and docstrings to use raysect's interpolators.

vsnever:
The docstrings of the Cherab interpolators are now updated. All demos in the `demos` directory have been updated to use Raysect interpolators, however Cherab interpolators are used in a couple of legacy CXRS demos in https://github.com/cherab/core/tree/development/docs/source/demonstrations/jet_cxrs that I left out. These CXRS demos require a significant update, not just replacing the interpolators.

The interpolators are also replaced in https://github.com/cherab/core/blob/22996e43f0a96f0633aeef145e74a837d043f491/cherab/tools/tests/test_ionization_balance.py. This test was commented out in Aug 2019 because it requires `cherab.openadas`, which was not part of the Cherab back than. Now the test runs fine, so I uncommented the code.

Also, there is a commented out file, https://github.com/cherab/core/blob/development/cherab/tools/equilibrium/equ.py, with references to the Cherab interpolators, that I have not touched. Is it safe to remove this file?

jacklovell:
> Also, there is a commented out file, https://github.com/cherab/core/blob/development/cherab/tools/equilibrium/equ.py, with references to the Cherab interpolators, that I have not touched. Is it safe to remove this file?

Yeah, let's remove it, as well as the commented out line in `__init__.py` which would have imported it. This code was commented out 2 releases ago and hasn't been touched since, so likely isn't in high demand. And I'd rather not have dead code lying around in the repository.

vsnever:
The tests fail due to `test_ionization_balance.py` (now uncommented) requires ionisation rates, which are missing because the OpenADAS repository is not populated. I think we can just add required json files to cherab/tools/tests/data and specify the path to repository in the test.

jacklovell:
I'm happy to have the JSON files kept in the respository as static test data. But I think we should get written permission from OpenADAS to do this, as we're effectively incorporating the data into Cherab in this instance. The [OpenADAS terms and conditions](https://open.adas.ac.uk/terms-and-conditions) state that we should seek permission for this. Could you contact them and check they're happy with this?

If permission is not forthcoming then we'll need to mock up some data files with dummy data and adjust the test as appropriate.

vsnever:
>  
> I'm happy to have the JSON files kept in the respository as static test data. But I think we should get written permission from OpenADAS to do this, as we're effectively incorporating the data into Cherab in this instance. The [OpenADAS terms and conditions](https://open.adas.ac.uk/terms-and-conditions) state that we should seek permission for this. Could you contact them and check they're happy with this?
> 
> If permission is not forthcoming then we'll need to mock up some data files with dummy data and adjust the test as appropriate.

Thanks for noticing this. I emailed Prof. Hugh Summers (you are in Cc), so let's wait for the answer. I agree that for the post-installation tests we should rather mock the data in case of negative response, but keep it locally, rather than downloading it from OPEN-ADAS at runtime.

CnlPepper:
I'd recommend using dummy data for the tests regardless of the answer. This is mainly to avoid potential future arguments about ownership when permission is inevitably forgotten about (any claims of illegality would fall on the individuals contributing).

vsnever:
The OPEN-ADAS data is replaced with mock-up data for the ionization balance test. 

The CI builds fail because the development branch of Raysect now has breaking changes in material API. PR #321 temporarily forces CI to build against Raysect master branch, which is now updated to 0.7.1.

jacklovell:
I've checked this PR out locally and I'm getting a bunch of errors from the ionization balance tests. For example:

```
======================================================================
ERROR: test_balance_2d_interpolators_from_element_density_tcx (cherab.tools.tests.test_ionization_balance.TestIonizationBalance1D)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/jlovell/cherab/core/cherab/tools/tests/test_ionization_balance.py", line 801, in test_balance_2d_interpolators_from_element_density_tcx
    tcx_donor_charge=0)
  File "/tmp/jlovell/cherab/core/cherab/tools/plasmas/ionisation_balance.py", line 757, in interpolators2d_from_elementdensity
    free_variable=free_variable)
  File "/tmp/jlovell/cherab/core/cherab/tools/plasmas/ionisation_balance.py", line 507, in from_elementdensity
    tcx_donor_n, tcx_donor_charge)
  File "/tmp/jlovell/cherab/core/cherab/tools/plasmas/ionisation_balance.py", line 475, in _from_elementdensity
    coef_ion, coef_recom, coef_tcx)
  File "/tmp/jlovell/cherab/core/cherab/tools/plasmas/ionisation_balance.py", line 286, in _from_element_density_point
    tcx_donor_n)
  File "/tmp/jlovell/cherab/core/cherab/tools/plasmas/ionisation_balance.py", line 240, in _fractional_abundance_point
    abundance = lsq_linear(matbal, rhs, bounds=(0, n_e))["x"]
  File "/tmp/jlovell/Cherab-PR-venv/lib/python3.7/site-packages/scipy/optimize/_lsq/lsq_linear.py", line 275, in lsq_linear
    x_lsq = np.linalg.lstsq(A, b, rcond=-1)[0]
  File "<__array_function__ internals>", line 6, in lstsq
  File "/tmp/jlovell/Cherab-PR-venv/lib/python3.7/site-packages/numpy/linalg/linalg.py", line 2306, in lstsq
    x, resids, rank, s = gufunc(a, b, rcond, signature=signature, extobj=extobj)
ValueError: On entry to DLASCL parameter number 4 had an illegal value
```
There are 36 tests which fail with this error, and another 6 which fail the comparison test.

vsnever:
> I've checked this PR out locally and I'm getting a bunch of errors from the ionization balance tests.

These tests worked fine with older versions of Numpy and Scipy. Now that I have upgraded to Numpy 1.21 and Scipy 1.7.2, I am getting the same errors. It looks like I need to be more accurate with mocking up the ADAS rates.

Interestingly, when using real OpenADAS data, `test_fractional_2d_from_mixed_tcx()` just hangs while all other 41 tests pass.

vsnever:
I updated the mockup atomic rates and the ionization balance tests now work correctly. However, I do not yet know what caused the interpolators to return incorrect values with old data. I'll try to figure it out.

jacklovell:
Thanks. If it's just a numerical precision issue then I don't think we need to worry about it. Let me know when you've finished investigating it, then I can merge this PR.

vsnever:
Looks like the tests failed with the old atomic data because Raysect's cubic interpolator should be used with caution on coarse grids. Here is an example:
```python
import numpy as np
from raysect.core.math.function.float import Interpolator1DArray
from cherab.core.math import Interpolate1DCubic
from matplotlib import pyplot as plt

x = np.array([0.2, 0.5, 1., 5., 10, 50., 100., 1000, 10000])
f = np.array([4.8e-43, 6.9e-26, 3.4e-20, 1.7e-15, 7.7e-15, 3.2e-14, 3.6e-14, 2.5e-14, 1.1e-14])

f_raysect = Interpolator1DArray(x, np.log10(f), 'cubic', 'none', np.inf)
f_cherab = Interpolate1DCubic(x, np.log10(f), continuity_order=1)
x_new = np.geomspace(0.21, 10000, 1024)
f_raysect_new = np.array([10**f_raysect(x1) for x1 in x_new])
f_cherab_new = np.array([10**f_cherab(x1) for x1 in x_new])
plt.plot(x_new, f_raysect_new, label='Raysect cubic interpolator')
plt.plot(x_new, f_cherab_new, label='Cherab cubic interpolator')
plt.legend()
plt.gca().set_xscale('log')
plt.gca().set_yscale('log')
plt.ylabel('Hydrogen ionization rate at Ne=1e19, m3 s-1')
plt.xlabel('Te, eV')
plt.show()
```
raysect_cherab_cubic_interp

On even coarser grids, the interpolated rate can easily go infinite, while on a finer grid from the Open-ADAS files, the Raysect and Cherab interpolators produce the same results.

However, I think that this problem could be avoided if the atomic rates were interpolated on a log-log scale rather than a semi-log scale. Currently, for the ionisation rate we have:
```python
cdef class IonisationRate(CoreIonisationRate):

    def __init__(self, dict data, extrapolate=False):

        self.raw_data = data

        # unpack
        ne = data['ne']
        te = data['te']
        rate = np.log10(data['rate'])

        # store limits of data
        self.density_range = ne.min(), ne.max()
        self.temperature_range = te.min(), te.max()

        # interpolate rate
        extrapolation_type = 'linear' if extrapolate else 'none'
        self._rate = Interpolator2DArray(ne, te, rate, 'cubic', extrapolation_type, INFINITY, INFINITY)

    cpdef double evaluate(self, double density, double temperature) except? -1e999:

        # calculate rate and convert from log10 space to linear space
        return 10 ** self._rate.evaluate(density, temperature)
```
Changing this to
```python
cdef class IonisationRate(CoreIonisationRate):

    def __init__(self, dict data, extrapolate=False):

        self.raw_data = data

        # unpack
        ne = data['ne']
        te = data['te']
        rate = np.log10(data['rate'])

        # store limits of data
        self.density_range = ne.min(), ne.max()
        self.temperature_range = te.min(), te.max()

        # moving to log scale
        ne = np.log10(ne)
        te = np.log10(te)

        # interpolate rate
        extrapolation_type = 'linear' if extrapolate else 'none'
        self._rate = Interpolator2DArray(ne, te, rate, 'cubic', extrapolation_type, INFINITY, INFINITY)

    cpdef double evaluate(self, double density, double temperature) except? -1e999:

        # calculate rate and convert from log10 space to linear space
        return 10 ** self._rate.evaluate(log10(density), log10(temperature))
```
should fix the problem with oscillating cubic splines on coarse grids, as shown in the following example:
```python
import numpy as np
from raysect.core.math.function.float import Interpolator1DArray
from cherab.core.math import Interpolate1DCubic
from matplotlib import pyplot as plt

x = np.array([0.2, 0.5, 1., 5., 10, 50., 100., 1000, 10000])
f = np.array([4.8e-43, 6.9e-26, 3.4e-20, 1.7e-15, 7.7e-15, 3.2e-14, 3.6e-14, 2.5e-14, 1.1e-14])

f_raysect_log_log = Interpolator1DArray(np.log10(x), np.log10(f), 'cubic', 'none', np.inf)
f_cherab = Interpolate1DCubic(x, np.log10(f), continuity_order=1)
f_cherab_log_log = Interpolate1DCubic(np.log10(x), np.log10(f), continuity_order=1)
x_new = np.geomspace(0.21, 10000, 1024)
f_raysect_log_log_new = np.array([10**f_raysect_log_log(np.log10(x1)) for x1 in x_new])
f_cherab_new = np.array([10**f_cherab(x1) for x1 in x_new])
f_cherab_log_log_new = np.array([10**f_cherab_log_log(np.log10(x1)) for x1 in x_new])
plt.plot(x_new, f_raysect_log_log_new, label='Raysect cubic interpolator (log-log)')
plt.plot(x_new, f_cherab_new, label='Cherab cubic interpolator (semi-log)')
plt.plot(x_new, f_cherab_log_log_new, label='Cherab cubic interpolator (log-log)')
plt.legend()
plt.gca().set_xscale('log')
plt.gca().set_yscale('log')
plt.ylabel('Hydrogen ionization rate, m3 s-1')
plt.xlabel('Te, eV')
plt.show()
```
raysect_cherab_cubic_interp_double_log

@jacklovell, @Mateasek, @CnlPepper, what are your thoughts on this? Should we replace semi-log interpolation with log-log interpolation for atomic rates? If so, then I'll create a separate issue for this.

CnlPepper:
This looks like a bug... it shouldn't do that.

vsnever:
> This looks like a bug... it shouldn't do that.

Actually, I was surprised when I saw this picture, because at some early stage of development of Raysect interpolators, I compared the analytical expressions for the cubic spline coefficients in Raysect and Cherab in 1D case, and found that they are the same except for the bounding cells.

CnlPepper:
Where are the knot points in the semi log graph? I'm trying to understand how that large bump has appeared. Moving to log-log might be a better approximation for low resolution rates. but I think it is actually masking an issue here.

CnlPepper:
Sorry was quickly scanning the thread and didn't pay attention to the code.... will run it and have a look where the difference is. I have a suspicion... it might be the differential calculation at each knot. If there is an error or difference in behaviour it is likely there. Will spend a few mins digging through it, but I can't afford mush time atm.

CnlPepper:
So just confirmed it is the differential calculation. The raysect interpolator uses a higher order expansion to calculate the gradient.... but I'm questioning if that was the right choice. Cherab uses the average gradient either side of the knot. My intuition lends to the average being more stable... born out by the quality of the interpolation. I'll have a chat with Simon about the implementation and consider if we should switch to the average.

CnlPepper:
So I've spent some more time mulling over this. The switch to the higher order expansion is the correct one. Highly irregular grids (as in the example code) should weight the differential according to the proximity of the neighbouring points. Cherab's code is incorrect here, Raysect is correct.

Could you plot a range of rates and check the interpolation is stable for representative data? If not, it may be necessary to interpolate in log-log space (might be worth doing this anyway to reduce the risk of poorly sampled data causing issues).

vsnever:
> Could you plot a range of rates and check the interpolation is stable for representative data?

Considering that atomic rates are two-dimensional, and the number of elements and ions is large enough, I'll better do some automated testing, probably using linear interpolation on log-log scale as a reference one.

vsnever:
I've scanned all the ionisation and recombination rates obtained with `populate()`. Here are a couple of the worst cases:
lithium1_ionisation
noen5_ionisation
Recombination rates are also sometimes interpolated with noticeable errors.
lithium2_recombination
My favorite is the log-log interpolation, however, it's ~10% slower than semi-log.
I haven't checked PECs yet. Cherab interpolates them on linear scale, I think the log-log interpolation can give a more accurate result.

vsnever:
I think we should switch to log-log interpolation for PECs. These PECs are for H-alpha:
hydrogen_3_2_impact_excitation_pec
hydrogen_3_2_recombination_pec
The region Te < 10 eV at Ne > 10<sup>20</sup> m<sup>-3</sup> is important because these conditions are typical for divertor plasma. Impact excitation PEC is small there but the density of neutral atoms is huge.

However, H-alpha is not the worst case. This is the recombination PEC for Be<sup>3+</sup> 2-->1 transition:
beryllium_3_2_1_recombination_pec

CnlPepper:
Agreed it looks like it will be necessary given the new differential calculation.

CnlPepper:
Thanks for plotting these, I was suspicious the data spacing combined with the new differential would cause occasional deviations. Some of them look a little extreme... are there two almost coincident points driving in the extreme gradients?

vsnever:
> Some of them look a little extreme... are there two almost coincident points driving in the extreme gradients?

I would not say that the Te points coincide, for the last plot the values are 4.14 eV and 6.89 eV, but the rate drops almost by an order of magnitude on this interval. Considering that the interpolation of the PECs is not even semi-log, the derivative can take extreme values. Here is the code for the last plot so you can check the details if you like. 

```python
import numpy as np
from matplotlib import pyplot as plt

from raysect.core.math.function.float import Interpolator2DArray
from cherab.openadas import OpenADAS
from cherab.core.atomic import hydrogen, beryllium
from cherab.core.math import sample2d_grid
from cherab.core.utility.conversion import PhotonToJ

atomic_data = OpenADAS()

ion = beryllium
charge = 3
transition = (2, 1)
ne_test = 5.e23

cherab_interp = atomic_data.recombination_pec(ion, charge, transition)
ne_min, ne_max = cherab_interp.density_range
te_min, te_max = cherab_interp.temperature_range
ne0 = cherab_interp.raw_data['ne']
te0 = cherab_interp.raw_data['te']
rate0 = PhotonToJ.to(cherab_interp.raw_data['rate'], cherab_interp.wavelength)
raysect_interp = Interpolator2DArray(ne0, te0, rate0, 'cubic', 'none', np.inf, np.inf)
raysect_interp_log_log = Interpolator2DArray(np.log10(ne0), np.log10(te0), np.log10(rate0), 'cubic', 'none', np.inf, np.inf)

ne_indx = np.argmin(np.abs(ne_test - ne0))
ne_point = ne0[ne_indx]
te = np.geomspace(te_min, te_max, 512)
cherab_rate = sample2d_grid(cherab_interp, ne_point, te)[0]
raysect_rate = sample2d_grid(raysect_interp, ne_point, te)[0]
raysect_log_log_rate = 10**sample2d_grid(raysect_interp_log_log, np.log10(ne_point), np.log10(te))[0]

fig = plt.figure()
ax = fig.add_subplot()
ax.plot(te, cherab_rate, label='Cherab interpolator')
ax.plot(te, raysect_rate, label='Raysect interpolator')
ax.plot(te, raysect_log_log_rate, label='Raysect log-log interpolator')
ax.plot(te0, rate0[ne_indx], 'o', mfc='none', mec='k')
ax.set_title('{} {}+ {}->{} recombination PEC at Ne = {} m-3'.format(ion.name, charge, transition[0], transition[1], ne_point))
ax.set_xscale('log')
ax.set_yscale('log')
ax.set_xlabel('Te, eV')
ax.set_ylabel('W m3 s-1')
ax.legend()
plt.show()
```

CnlPepper:
Thanks Vlad. @mattngc will probably chuckle as I originally removed log-log interpolation as it was unnecessary for the data sets we originally used. Look like it is now essential.

vsnever:
Checked `BeamStoppingRate`, `BeamPopulationRate` and `BeamEmissionPEC`. It's ok to use cubic interpolation in log-log space for them. 

Switching to cubic log-log is also fine for `ContinuumPower` and `CXRadiationPower`. The only rate that can cause problems in some rare cases is `LineRadiationPower`. Here is one of the worst cases:
carbon_5_line_radiated_power
I suggest using cubic interpolation anyway, as it's better than linear at Te > 20 eV. Also, such behaviour is quite rare.

`BeamCXPEC` is the only rate I have not checked yet.

vsnever:
I'm afraid, we have to switch to nearest extrapolation if we interpolate in log-log space, otherwise we'll get infinite values at zero temperature and density for some rates.
```python
import numpy as np
from matplotlib import pyplot as plt

from raysect.core.math.function.float import Interpolator2DArray
from cherab.openadas import OpenADAS
from cherab.core.atomic import lithium, oxygen
from cherab.core.math import sample2d_grid

atomic_data = OpenADAS(permit_extrapolation=True)

ion = lithium
charge = 1
ne_test = 1.e19
te_test = 0.2

cherab_interp = atomic_data.ionisation_rate(ion, charge)
ne_min, ne_max = cherab_interp.density_range
te_min, te_max = cherab_interp.temperature_range
ne0 = cherab_interp.raw_data['ne']
te0 = cherab_interp.raw_data['te']
rate0 = cherab_interp.raw_data['rate']
raysect_interp_log_log = Interpolator2DArray(np.log10(ne0), np.log10(te0), np.log10(rate0), 'cubic', 'linear', np.inf, np.inf)

ne_indx = np.argmin(np.abs(ne_test - ne0))
ne_point = ne0[ne_indx]
te = np.geomspace(1.e-10, te_max, 512)
cherab_rate = sample2d_grid(cherab_interp, ne_point, te)[0]
raysect_log_log_rate = 10**sample2d_grid(raysect_interp_log_log, np.log10(ne_point), np.log10(te))[0]

fig = plt.figure()
ax = fig.add_subplot()
ax.plot(te, cherab_rate, label='Cherab interpolator')
ax.plot(te, raysect_log_log_rate, label='Raysect log-log interpolator')
ax.plot(te0, rate0[ne_indx], 'o', mfc='none', mec='k')

ax.set_title('{} {}+ ionisation rate extrapolation at Ne = {} m-3'.format(ion.name, charge, ne_point))
ax.set_xscale('log')
ax.set_yscale('log')
ax.set_xlabel('Te, eV')
ax.set_ylabel('m3 s-1')
plt.legend()
plt.show()
```
lithium_1_ionisation_rate_extrapolation

CnlPepper:
Great. ADAS data strikes again....the problem with nearest is that there will be value where there shouldn't be. Really this should be fixed at the adas end.

If you make it nearest neighbour, you'll need to check the impact on the simulations that touch data just beyond the rate values. 

Everything about ADAS is infuriating. The community really needs to replace it with something more modern.

If the impact is low, you could temporarily use nearest. But I'd suggest we develop a system to address the awful data directly. Maybe adding points to the ends of the adas data to clamp the ridiculous gradients at yhe point of import (user selectable). The ADAS importing routines really need a "fixup" stage where we fix known issues with the data at yhe point of import into the cherab rate repo. This data gradient fix could be an example "fixup" for the rates with extrapolation issues. It would be substantially cleaner than hacking the core to handle ADAS's woeful data quality.

vsnever:
For now I'll switch to nearest extrapolation only where it's necessary: ionisation rate, recombination rate, excitation PEC and recombination PEC. All these rates are multiplied by electron density in the models, so if we have near-zero plasma density, the exact value of the rate doesn't matter much until it's low, and with nearest extrapolation it will be low. 

A combination of linear and quadratic extrapolation is fine for the beam rates, the linear extrapolation is fine for the thermal CX rates.

I still have to check the `BeamCXPEC` though.

By the way, I've checked Cherab's quadratic extrapolation, and in many cases the values at zero it gives are not significantly lower than the values at the low boundary of the ADAS data.

CnlPepper:
Thanks for exploring this in detail. Sounds like switching to nearest is the best solution near term. Would be worth updating the documentation for the rates to explain to users how the data is handled.

vsnever:
Checked the last rate, `BeamCXPEC`. `qeb` must be interpolated in log-log space, otherwise it oscillates. Quadratic extrapolation works well for `qeb`. Other quantities: `qti`, `qni` and `qzeff` can be interpolated in linear space, but the data for them is usually poor, so I switched to the nearest extrapolation. 

@jacklovell, I think this PR is ready.

@Mateasek, if you have workflows for plasma-beam interactions, please check them out with these new rates and see if you still get adequate results. I hope I didn't break anything.

PR #307, opened 2021-09-07
Add emitters.pxd and their __init__.pxd files

munechika-koyo:
Hello,

I would like to push this PR which allows external modules to handle the classes and functions in emitters.pyx.
I would appreciate it if the authors confirm it.

Thank you in advance.

Koyo


vsnever:
Hi @munechika-koyo, thanks for this. Looks fine to me.

PR #317, opened 2021-10-27
Improve metadata for nicer pypi page

dschwoerer:
Adding some links, to make the page at https://pypi.org/project/cherab/ look nicer.

I am not 100% certain the syntax is correct, I only tested it with setup.cfg ...


setup.py
```diff
@@ -97,6 +97,13 @@
         "Programming Language :: Python :: 3",
         "Topic :: Scientific/Engineering :: Physics"
     ],
+    url="https://github.com/cherab",
+    project_urls=dict(
+        Tracker="https://github.com/cherab/core/issues"
```
jacklovell, line 102: Syntax error: you're missing a comma

setup.py
```diff
@@ -97,6 +97,13 @@
...
+        Documentation="https://cherab.github.io/documentation/"
+    ),
+    long_description="file: README.md",
```
jacklovell, line 105: This syntax is lifted from setup.cfg, which is declarative rather than programmatic. I think for setup.py you need to read the contents of the file into a string and pass that as long description:
```python
with open("README.md") as f:
    long_description = f.read()
```

jacklovell (changes requested):
Thanks for this @dschwoerer. Could you please submit the PR against the development branch, not master? We only ever make changes against development, which is then merged to master with each release.

jacklovell (approved):
Looks good to me. Just tried a test release and it certainly provides lots more useful information than previously: https://test.pypi.org/project/cherab/1.2.1.dev0/

PR #321, opened 2021-11-07
Temporarily fix ci.yml to build Cherab against the released Raysect 0.7.1 regardless of branch.

### LINKED ISSUE #320: Cherab development branch is incompatible with Raysect development branch due to material API changes
Since Raysect 0.7.1 is released, I think until the new version of Cherab is ready we should set ci.yml to build with Raysect master branch.

vsnever:
This should fix #320 until the new version of Cherab is released and the development branch is updated to support Raysect 0.8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,12 +27,8 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install Python dependencies
       run: python -m pip install cython>=0.28 numpy==${{ matrix.numpy-version }} scipy matplotlib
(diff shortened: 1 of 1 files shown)


jacklovell:
Can you simplify this a bit to take out the `if` statement completely? Just leave the branch which installs raysect from PyPI. We can add in the branch later depending on the discussion in #320 if necessary, but at the moment this CI file produces misleading output "installing raysect from Github" (when it isn't) and won't install raysect at all if it ends up on the master branch. Then I'm happy to merge it.

vsnever:
Done.

Issue #322, opened 2021-11-11
Possible inconsistencies of origin and direction values of _SpectroscopicObserver0DBase 


Mateasek:
The [_SpectroscopicObserver0DBase](https://github.com/cherab/core/blob/dc74ba308c1e7a6ac179530ca8092b2e0d9f2e88/cherab/tools/observers/spectroscopy/base.py#L26) uses [origin](https://github.com/cherab/core/blob/dc74ba308c1e7a6ac179530ca8092b2e0d9f2e88/cherab/tools/observers/spectroscopy/base.py#L49) and [direction](https://github.com/cherab/core/blob/dc74ba308c1e7a6ac179530ca8092b2e0d9f2e88/cherab/tools/observers/spectroscopy/base.py#L66) properties to calculate the transform of an observer and keeps the values. This can lead to inconsistencies between the actual transform of the observer and saved values of origin and direction, when the transform is changed directly.

Instead of keeping the values in _origin and _direction I propose changing the property getters to calculate the value from the transform:

``` python
    @property
    def direction(self):
        # The observation direction of the sight line.
        return Vector3D(0, 0, 1).transform(self.transform)

        @property
    def origin(self):
        # The origin point of the sight line.
        return Point3D(0, 0, 0).transform(self.transform)
```
I will be happy to apply this, but since it is work of @vsnever I would like to hear is opinion

Mateasek:
At the end I needed the fix now for my work, @vsnever please feel free to accept/reject it.

vsnever:
Fixed in #323.

PR #323, opened 2021-11-11
Fix possible inconsistency in origin and direction

Mateasek:
Inconsistency could be achieved if the transform was changed directly
without changing origin and direction throught setters.

origin and direction getters calculate values of origin and direction
from transform, values are not kept in _origin/_direction.

related to issue #322

### DIFF
--- a/cherab/tools/observers/spectroscopy/base.py
+++ b/cherab/tools/observers/spectroscopy/base.py
@@ -43,24 +43,24 @@ class _SpectroscopicObserver0DBase:
     @property
     def origin(self):
         # The origin point of the sight line.
-        return self._origin
+        return Point3D(0, 0, 0).transform(self.transform)
 
     @origin.setter
     def origin(self, value):
         if not isinstance(value, Point3D):
             raise TypeError("Attribute 'origin' must be of type Point3D.")
-
-        if self._direction.x != 0 or self._direction.y != 0 or self._direction.z != 1:
+        
+        direction = self.direction
+        if direction.x != 0 or direction.y != 0 or direction.z != 1:
             up = Vector3D(0, 0, 1)
         else:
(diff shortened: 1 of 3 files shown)


vsnever:
Thanks for this fix, @Mateasek. The fix is fine, but could you also update initialization of `SpectroscopicSightLine` and `SpectroscopicFibreOptic`? Attributes `_origin` and `_direction` must be omitted and the respective parameters should be optional as you suggested in #324. So, for the `SpectroscopicSightLine` the initialization will look like:
```python
    def __init__(self, origin=None, direction=None, pipelines=None, parent=None, name=None):

        pipelines = pipelines or [SpectralRadiancePipeline0D(accumulate=False)]

        super().__init__(pipelines=pipelines, parent=parent, name=name)

        self.origin = origin or Point3D(0, 0, 0)
        self.direction = direction or Vector3D(0, 0, 1)
```

Mateasek:
@vsnever Do we need to set self.origin or self.direction if not passed? Wouldn't this be clearer?

``` python
        if origin is not None:
            self.origin = origin

        if direction is not None:
            self.direction = direction
```

If you don't pass origin and direction, the transform is not changed and you get the default behaviour, but I guess it is a minor difference.

vsnever:
> @vsnever Do we need to set self.origin or self.direction if not passed? Wouldn't this be clearer?
> 
> ```python
>         if origin is not None:
>             self.origin = origin
> 
>         if direction is not None:
>             self.direction = direction
> ```
> 
> If you don't pass origin and direction, the transform is not changed and you get the default behaviour, but I guess it is a minor difference.

This is ok too.

vsnever (approved):
I'm happy with this fix.

Mateasek:
cheers @vsnever

PR #328, opened 2021-11-20
Use isotope wavelengths instead of element ones when converting from photons/s to Watts in PECs for isotopes

### LINKED ISSUE #327: Use wavelengths for isotopes if available when converting from photons/s to Watts in PECs
The wavelengths used in `OpenADAS.*_pec()` functions are the element's and not the isotope's wavelengths:
https://github.com/cherab/core/blob/dc74ba308c1e7a6ac179530ca8092b2e0d9f2e88/cherab/openadas/openadas.py#L246-L267

These wavelength are only used to convert PECs from photons/s to Watts, so the error due to ignoring the isotope is somewhere in the 4th digit. Yet, I think it's better to fix this so as not to confuse users who might want to validate the PECs. After the change, the code for `OpenADAS.impact_excitation_pec()` will look like this:
```python
    def impact_excitation_pec(self, ion, charge, transition):
        
        ion_element = ion.element if isinstance(ion, Isotope) else ion

        try: 
            data = repository.get_pec_excitation_rate(ion_element, charge, transition, repository_path=self._data_path)

        except RuntimeError:
            if self._missing_rates_return_null:
                return NullImpactExcitationPEC()
            raise

        wavelength = self.wavelength(ion, charge, transition)

        return ImpactExcitationPEC(wavelength, data, extrapolate=self._permit_extrapolation)
```
Note that `wavelength = self.wavelength (ion, charge, transition)` is deliberately placed outside the try scope, because if the repository contains the PEC for that transition, but does not contain the wavelength, it is not a case of missing rate, but a case of data inconsistency.

vsnever:
This PR fixes #327 by replacing `repository.get_wavelength(ion.element, charge, transition)` calls with `self.wavelength(ion, charge, transition)` calls in `OpenADAS.*_pec()` functions.

### DIFF
--- a/cherab/openadas/openadas.py
+++ b/cherab/openadas/openadas.py
@@ -134,19 +134,19 @@ class OpenADAS(AtomicData):
         if isinstance(donor_ion, Isotope):
             donor_ion = donor_ion.element
 
-        if isinstance(receiver_ion, Isotope):
-            receiver_ion = receiver_ion.element
+        receiver_ion_element = receiver_ion.element if isinstance(receiver_ion, Isotope) else receiver_ion
 
         try:
             # read data
-            wavelength = repository.get_wavelength(receiver_ion, receiver_charge - 1, transition, repository_path=self._data_path)
-            data = repository.get_beam_cx_rates(donor_ion, receiver_ion, receiver_charge, transition,
+            data = repository.get_beam_cx_rates(donor_ion, receiver_ion_element, receiver_charge, transition,
                                                 repository_path=self._data_path)
 
         except RuntimeError:
             if self._missing_rates_return_null:
                 return [NullBeamCXPEC()]
             raise
+
+        wavelength = self.wavelength(receiver_ion, receiver_charge - 1, transition)
 
         # load and interpolate the relevant transition data from each file
         rates = []
@@ -223,22 +223,22 @@ class OpenADAS(AtomicData):
         """
 
         # extract element from isotope
-        if isinstance(beam_ion, Isotope):
-            beam_ion = beam_ion.element
+        beam_ion_element = beam_ion.element if isinstance(beam_ion, Isotope) else beam_ion
 
         if isinstance(plasma_ion, Isotope):
             plasma_ion = plasma_ion.element
 
         try:
             # locate data file
-            data = repository.get_beam_emission_rate(beam_ion, plasma_ion, charge, transition,
+            data = repository.get_beam_emission_rate(beam_ion_element, plasma_ion, charge, transition,
                                                      repository_path=self._data_path)
-            wavelength = repository.get_wavelength(beam_ion, 0, transition, repository_path=self._data_path)
 
         except RuntimeError:
             if self._missing_rates_return_null:
                 return NullBeamEmissionPEC()
             raise
+
+        wavelength = self.wavelength(beam_ion, charge, transition)
 
         # load and interpolate data
         return BeamEmissionPEC(data, wavelength, extrapolate=self._permit_extrapolation)
@@ -252,17 +252,17 @@ class OpenADAS(AtomicData):
         :return:
         """
 
-        if isinstance(ion, Isotope):
-            ion = ion.element
+        ion_element = ion.element if isinstance(ion, Isotope) else ion
 
         try:
-            wavelength = repository.get_wavelength(ion, charge, transition, repository_path=self._data_path)
-            data = repository.get_pec_excitation_rate(ion, charge, transition, repository_path=self._data_path)
+            data = repository.get_pec_excitation_rate(ion_element, charge, transition, repository_path=self._data_path)
 
         except RuntimeError:
             if self._missing_rates_return_null:
                 return NullImpactExcitationPEC()
             raise
+
+        wavelength = self.wavelength(ion, charge, transition)
 
         return ImpactExcitationPEC(wavelength, data, extrapolate=self._permit_extrapolation)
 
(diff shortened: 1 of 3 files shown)


cherab/openadas/openadas.py
```diff
@@ -134,20 +134,20 @@ def beam_cx_pec(self, donor_ion, receiver_ion, receiver_charge, transition):
         if isinstance(donor_ion, Isotope):
             donor_ion = donor_ion.element
 
-        if isinstance(receiver_ion, Isotope):
-            receiver_ion = receiver_ion.element
+        receiver_ion_element = receiver_ion.element if isinstance(receiver_ion, Isotope) else receiver_ion
```
Mateasek, line 137: Could you please add a comment that you are extracting here the element of the passed isotope of something of the sort? It would be great if the code flow was a little bit more commented, so later we have easier code reading.

cherab/openadas/openadas.py
```diff
@@ -134,20 +134,20 @@ def beam_cx_pec(self, donor_ion, receiver_ion, receiver_charge, transition):
...
 
         try:
             # read data
-            wavelength = repository.get_wavelength(receiver_ion, receiver_charge - 1, transition, repository_path=self._data_path)
-            data = repository.get_beam_cx_rates(donor_ion, receiver_ion, receiver_charge, transition,
+            data = repository.get_beam_cx_rates(donor_ion, receiver_ion_element, receiver_charge, transition,
```
Mateasek, line 141: Also here, could you please add a line comment saying the code is getting the element's data, because adas doesn't have isotope data?

cherab/openadas/openadas.py
```diff
@@ -134,20 +134,20 @@ def beam_cx_pec(self, donor_ion, receiver_ion, receiver_charge, transition):
...
                                                 repository_path=self._data_path)
 
         except RuntimeError:
             if self._missing_rates_return_null:
                 return [NullBeamCXPEC()]
             raise
 
+        wavelength = self.wavelength(receiver_ion, receiver_charge - 1, transition)
```
Mateasek, line 149: And here please a comment that it is the isotope's rest wavelength obtained, to have correct line shapes and radiated powers.

cherab/openadas/openadas.py
```diff
@@ -223,23 +223,23 @@ def beam_emission_pec(self, beam_ion, plasma_ion, charge, transition):
...
-            beam_ion = beam_ion.element
+        beam_ion_element = beam_ion.element if isinstance(beam_ion, Isotope) else beam_ion
 
         if isinstance(plasma_ion, Isotope):
             plasma_ion = plasma_ion.element
 
         try:
             # locate data file
-            data = repository.get_beam_emission_rate(beam_ion, plasma_ion, charge, transition,
+            data = repository.get_beam_emission_rate(beam_ion_element, plasma_ion, charge, transition,
```
Mateasek, line 233: as above

cherab/openadas/openadas.py
```diff
@@ -223,23 +223,23 @@ def beam_emission_pec(self, beam_ion, plasma_ion, charge, transition):
...
                                                      repository_path=self._data_path)
-            wavelength = repository.get_wavelength(beam_ion, 0, transition, repository_path=self._data_path)
 
         except RuntimeError:
             if self._missing_rates_return_null:
                 return NullBeamEmissionPEC()
             raise
 
+        wavelength = self.wavelength(beam_ion, 0, transition)
```
Mateasek, line 241: as above

cherab/openadas/openadas.py
```diff
@@ -252,18 +252,18 @@ def impact_excitation_pec(self, ion, charge, transition):
         :return:
         """
 
-        if isinstance(ion, Isotope):
-            ion = ion.element
+        ion_element = ion.element if isinstance(ion, Isotope) else ion
```
Mateasek, line 255: as above

cherab/openadas/openadas.py
```diff
@@ -252,18 +252,18 @@ def impact_excitation_pec(self, ion, charge, transition):
...
 
         try:
-            wavelength = repository.get_wavelength(ion, charge, transition, repository_path=self._data_path)
-            data = repository.get_pec_excitation_rate(ion, charge, transition, repository_path=self._data_path)
+            data = repository.get_pec_excitation_rate(ion_element, charge, transition, repository_path=self._data_path)
```
Mateasek, line 258: as above

cherab/openadas/openadas.py
```diff
@@ -252,18 +252,18 @@ def impact_excitation_pec(self, ion, charge, transition):
...
 
         except RuntimeError:
             if self._missing_rates_return_null:
                 return NullImpactExcitationPEC()
             raise
 
+        wavelength = self.wavelength(ion, charge, transition)
```
Mateasek, line 265: as above

cherab/openadas/openadas.py
```diff
@@ -275,18 +275,18 @@ def recombination_pec(self, ion, charge, transition):
         :return:
         """
 
-        if isinstance(ion, Isotope):
-            ion = ion.element
+        ion_element = ion.element if isinstance(ion, Isotope) else ion
```
Mateasek, line 278: as above

cherab/openadas/openadas.py
```diff
@@ -275,18 +275,18 @@ def recombination_pec(self, ion, charge, transition):
...
 
         try:
-            wavelength = repository.get_wavelength(ion, charge, transition, repository_path=self._data_path)
-            data = repository.get_pec_recombination_rate(ion, charge, transition, repository_path=self._data_path)
+            data = repository.get_pec_recombination_rate(ion_element, charge, transition, repository_path=self._data_path)
```
Mateasek, line 281: as above

cherab/openadas/openadas.py
```diff
@@ -275,18 +275,18 @@ def recombination_pec(self, ion, charge, transition):
...
 
         except (FileNotFoundError, KeyError):
             if self._missing_rates_return_null:
                 return NullRecombinationPEC()
             raise
 
+        wavelength = self.wavelength(ion, charge, transition)
```
Mateasek, line 288: as above

Mateasek:
Hi @vsnever , thanks for this fix. It is important to have it correct, no matter what % errors this fixed! Especially when the framework was ready for this. I had only a small remarks not related to the code. Adding some more explanatory comments will make it easier to read, but this is just my preference.

I tested the code for H/D alpha for all the radiation cases. Seems working to me.

During the review I noticed that the [BeamEmissionPEC](https://github.com/cherab/core/blob/dc74ba308c1e7a6ac179530ca8092b2e0d9f2e88/cherab/openadas/rates/beam.pyx#L154) is the only case without the wavelength attribute. Could you also add a fix for this issue in this PR? thanks

Mateasek (changes requested):

vsnever:
Hi @Mateasek, thanks for the review. I've added `wavelength` attribute to `BeamEmissionPEC` and improved the comments in the `OpenADAS` methods as you suggested.

Mateasek (approved):
Thank you for adding the comments and the wavelength attribute @vsnever. I think this PR can be accepted.

jacklovell (approved):
Looks fine to me. Good to have comments explaining the reasoning for the change within the code.

PR #329, opened 2021-11-22
Feature/pickle plasma models

Mateasek:
This PR makes plasma models pickle-able. The part of the changes in elements.pyx (`__hash__, __richcmp__`) are from PR #246


Mateasek:
This is ready to be reviewed.

jacklovell:
I'll let @CnlPepper cast his eye over this one, as I can't be certain whether your tests capture all the potential corner cases.

cherab/core/atomic/line.pyx
```diff
@@ -58,6 +58,16 @@ cdef class Line:
...
 
+    def __getstate__(self):
+        return {"element": self.element,
+                "charge": self.charge,
+                "transition": self.transition}
+    
+    def __setstate__(self, state):
+        self.element = state["element"]
+        self.charge = state["charge"]
+        self.transition = state["transition"]
```
vsnever, line 69: I think the default `__getstate__`,` __setstate__` will do the same. Is there a reason to override them?

cherab/core/model/plasma/recombination.pyx
```diff
@@ -47,6 +47,22 @@ cdef class RecombinationLine(PlasmaModel):
...
+        state["lineshape_kwargs"] = self._lineshape_kwargs
+        return state
+    
+    def __setstate__(self, state):
+        super().__setstate__(state)
+        self._line = state["line"]
+        self._lineshape_class = state["lineshape_class"]
+        self._lineshape_args = state["lineshape_args"]
+        self._lineshape_kwargs = state["lineshape_kwargs"]
+        self._change()
```
vsnever, line 64: This assumes that `lineshape_args` and `lineshape_kwargs` are pickle-able objects. For example, for `ZeemanMultiplet`  that's not true, because it initializes with the `ZeemanStructure` instance, which contains `Function1D` objects.

vsnever (commented):
Thanks for this PR, @Mateasek. My only concern is that this will not work with the `ZeemanMultiplet` and other possible future line shape models with non-pickle-able arguments (see the comments).

cherab/core/model/plasma/recombination.pyx
Mateasek: damn, I din't check the upcoming changes. Well I will look at what could be done, but I'm not optimistic :-/

vsnever: I think `Interpolator1DArray` will be used in `ZeemanStructure` in 99% of cases. It cannot be pickled now, but this can be fixed if needed. But there is no guarantee that any other child of `Function1D` will not be used.

Mateasek: This PR was also motivated by the increasing pressure on data provenance. This was supposed to help a bit. If it is not possible due to the accepted changes, lets not push it. The provenance can be achieved with other solutions.

vsnever: I was going to propose adding `is_pickleable` class attribute to `LineShapeModel`, so `__setstate__` could raise RuntimeError when trying to pickle a `PlasmaModel` with non-pickle-able line shape. Are you sure you want to close it?

PR #330, opened 2021-11-25
Enhancement/targeted pixel group

skuba31:
This PR adresses changes in group observers structure proposed in #325 
Added class Observer0DGroup for working with groups of 0D observers
Changed names of methods and attributes including "sight_line" to "observer"
Added subclasses for Pixel and TargetedPixel observers
Changed spectroscopic groups to be subclasses of new Observer0DGroup
Added sight_line named wrappers for new observer attributes and methods to Spectroscopic groups for backward compatibility

Issue #325, opened 2021-11-19
Add TargetedPixelGroup


skuba31:
Similar to the OpticalFiberGroup, this class could be used for photodiodes with pinholes or collimators. Typical use case would be AXUV photodiode arrays with pinhole that are often used as fast bolometers on tokamaks.

vsnever:
Thanks for the idea, @skuba31.

In terms of implementation, the current version of `Observer0DGroup` is actually `SpectroscopicObserver0DGroup` because it only works with `SpectroscopicObserver0D` from Cherab, so we will need to rename it. The new `Observer0DGroup` should work with basic 0D observers from Raysect. Then inheritance will be as follows:

`Observer0DGroup(Node)`
`SpectroscopicObserver0DGroup(Observer0DGroup)`
`FibreOpticGroup(SpectroscopicObserver0DGroup)`
`SightLineGroup(SpectroscopicObserver0DGroup)`
`TargetedPixelGroup(Observer0DGroup)`

Mateasek:
You could also include the regular pixel, which would make the non-imaging group complete, except the mesh observers, where this doesn't make much sense.

skuba31:
Thanks for the breakdown @vsnever . @Mateasek I will include PixelGroup as well.

jacklovell:
Fixed in #332

PR #332, opened 2021-11-26
Add group observers for Pixel and TargetedPixel

skuba31:
This PR adresses changes in group observers structure proposed in #325 reworked #330  to correctly include changes
Added class Observer0DGroup for working with groups of 0D observers
Changed names of methods and attributes including "sight_line" to "observer"
Added subclasses for Pixel and TargettedPixel observers
Changed spectroscopic groups to be subclasses of new Observer0DGroup
Added sight_line named wrappers for new observer attributes and methods to Spectroscopic groups for backward compatibility


jacklovell:
Thanks for the contribution @skuba31. It'll take some time to review this properly and we're busy getting ready for a new release at the moment, so I'll take a proper look at this after the release is made.

vsnever:
Hi @skuba31, thank you very much for this contribution and sorry for delayed response. 
I started reviewing the code and realised that it might be worth changing the concept a little.

After generalizing group observers to support basic 0D observers from Raysect, having special spectroscopic observers, namely `SpectroscopicFibreOptic` and `SpectroscopicSightLine`, in Cherab seems redundant to me. I think it would be simpler and more convenient if `SightLineGroup` and `FibreOpticGroup` supported the `SightLine` and `FibreOptic` observers from Raysect. Additional features of `SpectroscopicObserver0DGroup`, such as displaying the results of group observation and controlling the parameters of individual pipelines through group interfaces are applicable to all 0D observers and can be implemented in `Observer0DGroup`. To do this, the code from `_SpectroscopicObserver0DBase` must be moved to the corresponding functions/properties of `Observer0DGroup`. In this approach, we will have a base class `Observer0DGroup` with some additional functions and properties from `SpectroscopicObserver0DGroup` and 4 child classes: `PixelGroup`, `TargettedPixelGroup`, `SightLineGroup` and `FibreOpticGroup`, which use observers from Raysect.

The `origin` and `direction` properties of `_SpectroscopicObserver0DBase` do not simplify the user code in my opinion, because using `translate()` and `rotate_basis()` is just as easy, so I think these observers can be marked deprecated especially considering that specifying just the forward direction without the up direction will not work with polarised light as @Mateasek pointed out in another conversation.

@CnlPepper, @jacklovell, @mateasek, do you have any thoughts on this?

Mateasek:
I'm definitely in favor of moving towards this. I never really liked the idea of shifting from the transforms to the origin and direction. Yes, transform is a tiny bit harder to get around in head and to get right, but it is the key building block of Raysect. Let's leave this to be done by external codes and packages.

Regarding the inheritance you proposed, it sound more "natural" to me. 

The only drawback is that the proposed changed are backwards incompatible, aren't they?

How would you like to continue? Do it in this PR, or make a new one?

skuba31:
@vsnever thanks for the update. Option 3 is fine with me. Is there a deadline to be met?

vsnever:
Thank you very much, @skuba31. There is no deadline.

cherab/tools/observers/group/pixel.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @Observer0DGroup.observers.setter
+    def observers(self, value):
+
+        if not isinstance(value, (list, tuple)):
+            raise TypeError("The observers attribute of PixelGroup must be a list or tuple of Pixels.")
+
+        for sight_line in value:
+            if not isinstance(sight_line, Pixel):
+                raise TypeError("The sight_lines attribute of LineOfSightGroup must be a list or tuple of "
```
vsnever, line 47: "The observers attribute of PixelGroup must be a list or ..."

cherab/tools/observers/group/sightline.py
```diff
@@ -18,10 +18,11 @@
 # under the Licence.
 
 from cherab.tools.observers.spectroscopy import SpectroscopicSightLine
+from .spectroscopic import SpectroscopicObserver0DGroup
 from .base import Observer0DGroup
 
 
-class SightLineGroup(Observer0DGroup):
+class SightLineGroup(SpectroscopicObserver0DGroup):
```
vsnever, line 25: Similar to `FibreOpticGroup`, this class should inherit from `Observer0DGroup` and use `SightLine` fomr Raysect instead of `SpectroscopicSightLine` from Cherab.

cherab/tools/observers/group/base.py
```diff
@@ -83,420 +78,257 @@ def __getitem__(self, item):
...
-                sight_lines.append(sight_line)
-
-        if len(pipelines) == 0:
-            raise ValueError("Pipeline {} was not found for any sight-line in this {}.".format((item, self.__class__.__name__)))
-
-        pipeline_types = set(type(pipeline) for pipeline in pipelines)
-        if len(pipeline_types) > 1:
-            raise ValueError("Pipelines {} have different types for different sight-lines.".format(item))
-
-        return pipelines, sight_lines
```
vsnever, line 412: I missed this function in my review. I suggest it should be changed the following way:
```python
    def _get_same_pipelines(self, item):
        pipelines = []
        observers = []
        for observer in self._observers:
            if isinstance(item, int):
                if item < len(observer.pipelines):
                    pipelines.append(observer.pipelines[item])
                    observers.append(observer)
            elif isinstance(item, str):
                matching_pipelines = [pipeline for pipeline in observer.pipelines if pipeline.name == item]
                if len(matching_pipelines) == 1:
                    pipelines.append(matching_pipelines[0])
                    observers.append(observer)

        if len(pipelines) == 0:
            raise ValueError("Pipeline {} was not found for any observer in this {}.".format((item, self.__class__.__name__)))

        pipeline_types = set(type(pipeline) for pipeline in pipelines)
        if len(pipeline_types) > 1:
            raise ValueError("Pipelines {} have different types for different observers.".format(item))

        return pipelines, observers
```

cherab/tools/observers/group/base.py
```diff
@@ -18,33 +18,26 @@
...
-    :ivar list/Point3D origin: The origin points for the sight lines.
-    :ivar list/Vector3D direction: The observation directions for the sight lines.
+    :ivar list pipelines: A list of all pipelines connected to each observer in the group.
     :ivar list/RenderEngine render_engine: Rendering engine used by the sight lines.
                                            Note that if the engine is shared, changing its
                                            parameters for one sight line in a group will affect
                                            all sight lines.
-    :ivar list/bool display_progress: Toggles the display of live render progress.
-    :ivar list/bool accumulate: Toggles whether to accumulate samples with subsequent
-                                observations.
```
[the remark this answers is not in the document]
skuba31: This could be problematic as the pipeline base class does not implement display_progress or accumulate. This was ok for the SpectroscopicGroup which was inteded for spectroscopic pipelines which have these implemented.

cherab/tools/observers/group/base.py
```diff
@@ -83,420 +78,257 @@ def __getitem__(self, item):
...
         else:
             if not isinstance(value, RenderEngine):
                 raise TypeError("The list 'render_engine' must contain only RenderEngine instances.")
-            for sight_line in self._sight_lines:
-                sight_line.render_engine = value
-
-    @property
-    def display_progress(self):
-        # Toggles the display of live render progress.
-        return [sight_line.display_progress for sight_line in self._sight_lines]
```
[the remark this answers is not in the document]
skuba31: The base class for Observer0D does not have such functionality, therefore, I think it is not a good idea to have this property in the group base class.

cherab/tools/observers/group/base.py
```diff
@@ -83,420 +78,257 @@ def __getitem__(self, item):
...
-                                Filters are applied to the mono pipelines only, namely,
-                                PowerPipeline0D or RadiacnePipeline0D. The values provided for spectral
-                                pipelines will be ignored. The filter must be an instance of
-                                SpectralFunction or None.
-
-        """
-
-        for sight_line in self._sight_lines:
-            sight_line.connect_pipelines(properties)
-
```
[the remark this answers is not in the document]
skuba31: The pipelines property should definitely be included. I think that the readability of the connect_pipelines could be improved. Should I implement it "as is" or should I try to improve it?

cherab/tools/observers/group/base.py
```diff
@@ -83,420 +78,257 @@ def __getitem__(self, item):
...
-                ax.set_title('{}: pipeline {}'.format(self.name, item))
-        elif isinstance(item, str):
-            ax.set_title('{}: {}'.format(self.name, item))
-
-        ax.set_ylabel(ylabel)
-        ax.set_xlabel('Line of sight')
-
-        return ax
-
-    def plot_spectra(self, item=0, in_photons=False, ax=None):
```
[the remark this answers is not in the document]
skuba31: I think this method is not well suited for the base class as this class can be used with other than spectroscopic pipelines. It seems to me that in this case it would make sense to have a standalone function taking Observer0DGroup or list of pipelines as a parameter.

cherab/tools/observers/group/base.py
```diff
@@ -83,420 +78,257 @@ def __getitem__(self, item):
...
-        if len(pipelines) == 0:
-            raise ValueError("Pipeline {} was not found for any sight-line in this {}.".format((item, self.__class__.__name__)))
-
-        pipeline_types = set(type(pipeline) for pipeline in pipelines)
-        if len(pipeline_types) > 1:
-            raise ValueError("Pipelines {} have different types for different sight-lines.".format(item))
-
-        return pipelines, sight_lines
-
-    def plot_total_signal(self, item=0, ax=None):
```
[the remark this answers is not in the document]
skuba31: This method is definitely useful. I think it is worth consideration whether to include this visualisation/postprocessing method in a class which is inteded for control of observers or if it should be a standalone function in order to keep the base class smaller.

cherab/tools/observers/group/base.py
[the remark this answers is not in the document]
vsnever: Agreed. Let's remove this, it's not a problem to loop over the observer's pipelines and toggle these parameters manually if needed.

cherab/tools/observers/group/base.py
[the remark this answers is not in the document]
vsnever: Agreed.

cherab/tools/observers/group/base.py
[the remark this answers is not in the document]
vsnever: Yes, please improve this. Probably, `connect_pipelines()` should take a list of (`pipeline_class`, `pipeline_kwargs`) tuples. 

When doing the group observation, it will make sense having `display_progress` set to False in the default kwargs for the spectral pipelines.

cherab/tools/observers/group/base.py
vsnever: If plotting functions are moved outside the base class, this function must be public.

cherab/tools/observers/group/base.py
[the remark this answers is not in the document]
vsnever: Probably, moving this function and `plot_spectra()` outside the base class would do better. In this case they should be renamed to something like `plot_group_total()` and `plot_group_spectra()`.

cherab/tools/observers/group/base.py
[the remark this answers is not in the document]
skuba31: This seems as reasonable approach. But I am wondering if it would be possible to create interface so that group would be compatible with 1D pipelines, because the most common use case seems to be to have observers with the same pipelines.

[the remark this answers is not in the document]
vsnever: It is possible but you'll need to completely rewrite the `observe()` method. I'm not sure if this is a good approach. First, the current implementation allows different observers in the group have different spectral settings. If all 0D observers share the same 1D pipeline, this would not be possible. Second, I think this breaks the integrity. For example, you'll loose the observed frame, if you add a new observer to the group between observations even if the pipeline is in accumulation mode. 

However, if it is OK that all observers in a group have the same spectral properties and differ only in geometry, then the group observer can be implemented as `Observer1D`. I think it's easier than sharing 1D pipeline between multiple 0D observers. The disadvantage of this approach is that 1D observers do not have a `samples_per_task` property. Thus, if `MulticoreEngine` is used and the number of observers in the group is less than the number of parallel processes, rendering will not be efficiently parallelized.  Let's hear what @Mateasek, @jacklovell and @CnlPepper have to say about this.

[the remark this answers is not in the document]
vsnever: Actually, choosing between keeping as is or converting to `Observer1D`, I prefer to leave it as is.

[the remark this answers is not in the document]
skuba31: I see, let's keep it as it is now.

vsnever:
@skuba31, I would prefer to keep the `SpectroscopicFibreOptic` and `SpectroscopicSightLine` in the code for a couple of releases. Otherwise we break the user space without warning. Just add something like:
```python
"""
.. deprecated:: 1.4.0
       Use Raysect's FibreOptic observer instead.
""" 
```
in the beginning of their docstrings.

cherab/tools/observers/group/base.py
skuba31: I have moved this function into plotting module.

cherab/tools/observers/group/targettedpixel.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @x_width.setter
+    def x_width(self, value):
+        if isinstance(value, (list, tuple, ndarray)):
+            if len(value) == len(self._observers):
+                for pixel, v in zip(self._observers, value):
+                    pixel.x_width = v
+            else:
+                raise ValueError("The length of 'x_width' ({}) "
+                                 "mismatches the number of pixels ({}).".format(len(value), len(self._sight_lines)))
```
vsnever, line 80: Correct `len(self._sight_lines)` to `len(self._observers)`.

cherab/tools/observers/group/targettedpixel.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @y_width.setter
+    def y_width(self, value):
+        if isinstance(value, (list, tuple, ndarray)):
+            if len(value) == len(self._observers):
+                for pixel, v in zip(self._observers, value):
+                    pixel.y_width = v
+            else:
+                raise ValueError("The length of 'y_width' ({}) "
+                                 "mismatches the number of pixels ({}).".format(len(value), len(self._sight_lines)))
```
vsnever, line 97: As above.

cherab/tools/observers/group/targettedpixel.py
```diff
@@ -0,0 +1,137 @@
...
+    
+    @targetted_path_prob.setter
+    def targetted_path_prob(self, value):
+        if isinstance(value, (list, tuple)):
+            if len(value) == len(self._observers):
+                for pixel, v in zip(self._observers, value):
+                    pixel.targetted_path_prob = v
+            else:
+                raise ValueError("The length of 'value' ({}) "
+                                 "mismatches the number of pixels ({}).".format(len(value), len(self._sight_lines)))
```
vsnever, line 134: As above.

cherab/tools/observers/group/targettedpixel.py
```diff
@@ -0,0 +1,137 @@
...
+                for pixel, v in zip(self._observers, value):
+                    pixel.target = v
+                    v.parent = self
+            else:
+                raise ValueError("The length of 'value' ({}) "
+                                 "mismatches the number of pixels ({}).".format(len(value), len(self._sight_lines)))
+        else:
+            value.parent = self
+            for pixel in self._observers:
+                pixel.target = value
```
vsnever, line 120: The correct property name is `pixel.targets`, not `pixel.target`. `TargettedPixel` receives a list of targets, not a single target, so here we have to check if the supplied list contains lists/tuples or primitives. The target primitives must be connected to the same scene graph as the `TargettedPixel`, which means they must have the same root, but changing the parent of the target will affect coordinate space transformations, so the parent should not be changed. I propose the following implementation: 
```python
    @targets.setter
    def targets(self, value):
        if all(isinstance(v, (list, tuple)) for v in value):
            if len(value) == len(self._observers):
                for pixel, v in zip(self._observers, value):
                    pixel.targets = v
            else:
                raise ValueError("The length of 'value' ({}) "
                                 "mismatches the number of pixels ({}).".format(len(value), len(self._observers)))
        else:
            # assuming a list of primitives, the pixel's setter will throw an error if not
            for pixel in self._observers:
                pixel.targets = value
```

vsnever (changes requested):
Thank you @skuba31, I really like your implementation. I have one major comment on `TargettedPixelGroup` and a few minor correction to some docstrings and error messages.

Could you also add `TargettedPixelGroup` and `PixelGroup` to docs/source/tools/observers.rst?

cherab/tools/observers/group/base.py
```diff
@@ -18,33 +18,29 @@
...
-from raysect.optical import Spectrum
-from raysect.optical.observer import SpectralRadiancePipeline0D, SpectralPowerPipeline0D, RadiancePipeline0D
+from raysect.optical.observer import Observer0D
 
 
 class Observer0DGroup(Node):
     """
-    A base class for a group of 0D spectroscopic observers under a single scene-graph node.
+    .. deprecated:: 1.4.0
+       Use Raysect's observer classes instead.
```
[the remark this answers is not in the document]
skuba31: Moved the deprecated indication to the right base.py (with definition of _SpectroscopicObserver0DBase)

cherab/tools/observers/group/spectroscopic.py
```diff
@@ -0,0 +1,267 @@
...
+
+import matplotlib.pyplot as plt
+from numpy import ndarray
+
+from cherab.tools.observers.group.base import Observer0DGroup
+from raysect.optical import Spectrum
+from raysect.optical.observer import SpectralRadiancePipeline0D, SpectralPowerPipeline0D, RadiancePipeline0D
+
+
+class SpectroscopicObserver0DGroup(Observer0DGroup):
```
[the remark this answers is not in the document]
skuba31: I kept the class so that we can have SpectroscopicGroup for backward compatibility. Sublcasses were also added to this file so that all the depracated classes would be here.

skuba31:
Thanks @vsnever for the review. I have modified the code and the docs.

vsnever (approved):
Thanks @skuba31, now I'm happy with this PR.

cherab/tools/observers/group/plotting.py
```diff
@@ -0,0 +1,157 @@
...
+    :param Observer0DGroup group: Observer group from which to select pipelines.
+    :param str/int item: The index or name of the pipeline to be selected.
+    :return list pipelines: matching pipelines
+    :return list observers: observers with matching pipelines
+    """
+    pipelines = []
+    observers = []
+    for observer in group._observers:
+        if isinstance(item, int):
+            if item < len(observer.pipelines):
```
jacklovell, line 37: Raise an `IndexError` here: you can then explain exactly why no pipeline was found. You can either use this check or rely on the `IndexError` which would be returned when trying to access `observer.pipelines[item]`. Also, what happens if the index is negative? Is it what the user would expect? Especially if `abs(index) > len(observer.pipelines)`: you'll get an `IndexError` anyway.

cherab/tools/observers/group/plotting.py
```diff
@@ -0,0 +1,157 @@
...
+                pipelines.append(observer.pipelines[item])
+                observers.append(observer)
+        elif isinstance(item, str):
+            matching_pipelines = [pipeline for pipeline in observer.pipelines if pipeline.name == item]
+            if len(matching_pipelines) == 1:
```
jacklovell, line 42: Ff there is more than 1 matching pipeline, this check will silently fail. Best to invert the logic in the check: if there is not exactly 1 matching pipeline then raise an error.

cherab/tools/observers/group/plotting.py
```diff
@@ -0,0 +1,157 @@
...
+            spectrum = Spectrum(pipeline.min_wavelength, pipeline.max_wavelength, pipeline.bins)
+            spectrum.samples[:] = pipeline.samples.mean
+            signal.append(spectrum.total())
+        else:
+            signal.append(pipeline.value.mean)
+
+        if observer.name and len(observer.name):
+            tick_labels.append(observer.name)
+        else:
+            tick_labels.append(group._observers.index(observer))
```
jacklovell, line 83: Use `group.observers`

jacklovell (changes requested):
Thanks for this @skuba31. It's looking pretty good from my end: I agree with what @vsnever has already suggested and have only a few suggestions myself.

Could you add some tests for the new observers? It can be simple things like checking assignment and retrieval of properties, and that invalid inputs are correctly detected. Aim for as much coverage of the new classes as possible, within reason.

I notice there's quite a lot of repetition in each of the sub-classes, in particular in the `observers` setter and the `add_observer` method. The only thing that differs for each sub-class is the type of the observer to add: this can be parametrised by a class variable instead. Something like the following:
```python
class TargettedPixelGroup(Observer0DGroup):
    _OBSERVER_TYPE = TargettedPixel
    ...

class SightlineGroup(Observer0DGroup):
    _OBSERVER_TYPE = Sightline
    ...
```
Then the relevant part of the setter in `Observer0DGroup` becomes:
```python
@observers.setter
def observers(self, value):
    ...  # Check type of value as before
    for observer in value:
        if not isinstance(observer, self._OBSERVER_TYPE):
            raise TypeError("Observer should be of type {}".format(self._OBSERVER_TYPE))
    # etc.
```
You can then do the same with the `add_observer` method. Define both of these only in the base class, then if we need to add new functionality, new checks or bug fixes they only need to be done in one place.

jacklovell:
Oh, almost forgot. Can you add an entry to the changelog too for these new features? It'll be for version 1.4.0 (release date TBD).

cherab/tools/observers/group/plotting.py
vsnever: I agree with the comment on negative indexing, but I don't think that the `IndexError` should be raised. This function returns the lists of matching pipelines and the observers for which those pipelines are found. The observers in a group can have different number of pipelines, and if the _i_-th pipeline cannot be found for the _k_-th observer because this _k_-th observer has _N<sub>k</sub>_ < _i_ number of pipelines, then the _k_-th observer should be skipped because it has no matching pipeline.

cherab/tools/observers/group/plotting.py
vsnever: Partially agree, the error should be raised but only if `len(matching_pipelines) > 1`, because in this case we don't know which pipeline to choose. If there is no matching pipeline for this observer, it should be skipped.

cherab/tools/observers/group/plotting.py
```diff
@@ -0,0 +1,157 @@
...
+            # turn the samples into photon/s
+            spectrum = spectrum_observed.new_spectrum()
+            spectrum.samples[:] = spectrum_observed.to_photons()
+        else:
+            spectrum = spectrum_observed
+
+        if spectrum.samples.size > 1:
+            ax.plot(spectrum.wavelengths, spectrum.samples, label=observer.name)
+        else:
+            ax.plot(spectrum.wavelengths, spectrum.samples, marker='o', ls='none', label=observer.name)
```
vsnever, line 137: Sorry, I missed this. I think similarly to `plot_group_total()` we should check if the name of the observer is set, and if not, use its index as a label.
```python
label = observer.name if observer.name and len(observer.name) else group.observers.index(observer)
```

cherab/tools/observers/group/plotting.py
skuba31: I have added a warning if invalid index is used.

cherab/tools/observers/group/plotting.py
skuba31: I have added a warning if no matching pipeline was found. An error is raised only if there are more pipelines.

skuba31:
@jacklovell I have changed the observers setter and add_observer methods. The tests might take a bit longer as this is something I have little experience with and I might not have enougt time in the next few weeks.  
Regarding the changelog, I will have to synchronize with the new development branch as this PR started with older version. Should I reset the history and create a new PR after that, or shoul I keep this one?

jacklovell:
It's fine to keep working in this branch. Just merge the changes from development into your branch. Something like this, assuming your `enhancement/TargetedPixel` branch is the current working branch:
```bash
git fetch origin development:development
git merge development
```
Best to do this before adding your changelog entry, otherwise you'll have merge conflicts to resolve. 

If you're not confident doing this then just add the changelog anyway, and I'll handle the conflicts and re-sync with the development branch when the pull request is merged. It shouldn't be hard to do, as development and your branch haven't diverged significantly.

Regarding tests, have a look at some of the examples in cherab/tools/tests: they should give you an idea. I don't think we specifically have tests for other observers yet (I should probably add some for the bolometry...), but something like `test_plasma_slabs` is a good example.

skuba31:
@jacklovell I have merged the development branch using squash commit and added the unit tests for groups.

jacklovell:
It would have been better just to do a normal merge (not squashed) of development, as the squash commit seems to have sent Github mad and added a huge number of changes to the PR. It's not a big problem as the change set is sane when I check it out locally, and it'll clean up once it's merged. Unit tests look promising: I'll take a proper look at this next week but assuming no major surprises it should be good to merge.

jacklovell:
@skuba31 the tests look good: coverage is excellent and it seems they highlighted a few small bugs which are now fixed. I see there's no test coverage for the plotting though. This is fair enough as it's difficult to test plotting with automated unit tests. I think a demo would be good to add, both to illustrate how to do the plotting and also as a quick check that the functions work. You can put an example in `demos/observers`. Have a look at the existing demos for inspiration.

vsnever:
Hi @skuba31, I think the easiest way to create the demo requested by @jacklovell is to use the [generomak_raw_files_solps_line_emitter.py](https://github.com/cherab/solps/blob/master/demos/generomak_raw_files_solps_line_emitter.py) demo from Cherab-SOLPS. Just replace:
```python
from cherab.solps import load_solps_from_raw_output
...
demos_directory = os.path.dirname(__file__)
simulation_directory = os.path.join(demos_directory, 'data', 'raw')
print('Loading simulation...')
sim = load_solps_from_raw_output(simulation_directory)
print('Creating plasma...')
plasma = sim.create_plasma()
```
with
```python
from cherab.generomak.plasma import get_edge_plasma
...
plasma = get_edge_plasma()
```
remove this caching:
```python
# Caching the H-alpha radiation power density on the SOLPS mesh
min_wavelength = 654.
max_wavelength = 658.
print('Caching H-alpha radiation power density on SOLPS mesh...')
direction = Vector3D(0, 0, 1)
halpha_radiation = np.zeros_like(sim.electron_density)  # total radiation density on SOLPS mesh
spectrum = Spectrum(min_wavelength, max_wavelength, 1)
for ix in range(halpha_radiation.shape[1]):
    for iy in range(halpha_radiation.shape[0]):
        point = Point3D(sim.mesh.cr[iy, ix], 0, sim.mesh.cz[iy, ix])
        for model in plasma.models:
            halpha_radiation[iy, ix] += 4. * np.pi * model.emission(point, direction, spectrum.new_spectrum()).total()

sim.halpha_total_radiation = halpha_radiation

# Replace plasma models with the SOLPSLineEmitter
plasma.models = [SOLPSLineEmitter(sim.halpha_total_radiation_f2d, h_alpha)]
```
and this plotting:
```python
# Plotting the projection of sight lines on the poloidal plane.
plot_emission_and_sightlines(sim, group)
```
Then update the respective code for the deprecated `FibreOpticGroup` with the code for the new `FibreOpticGroup`. 

I'll update the `generomak_raw_files_solps_line_emitter.py` to use the new `FibreOpticGroup` once this PR is merged.

skuba31:
I have added the demo for group observers. But I had to make changes in [cherab.generomak.plasma.plasma.py](https://github.com/cherab/core/blob/0eb7454137b17bfbeca6d5e3eae4281f9cd4c157/cherab/generomak/plasma/plasma.py)
[line 63](https://github.com/cherab/core/blob/0eb7454137b17bfbeca6d5e3eae4281f9cd4c157/cherab/generomak/plasma/plasma.py#L63)
```python
profiles_dir = os.path.join(os.path.dirname(__file__), "/data/plasma/edge")
```
to
```python
profiles_dir = os.path.join(os.path.dirname(__file__), "data/edge")
```

[line 178](https://github.com/cherab/core/blob/0eb7454137b17bfbeca6d5e3eae4281f9cd4c157/cherab/generomak/plasma/plasma.py#L178)
```python
profiles_dir = os.path.join(os.path.dirname(__file__), "data/plasma/edge")
```
to
```python
profiles_dir = os.path.join(os.path.dirname(__file__), "data/edge")
```

Should I include these changes in this PR as well?

vsnever:
> Should I include these changes in this PR as well?

Yes please do it. We argued about where to put the data in https://github.com/cherab/core/pull/337, so it's not surprising that we ended up missing this. Thanks for the demo, it's the first demo to use the Generomak edge plasma.

skuba31:
@vsnever It should be fixed now.

vsnever:
@jacklovell, don't merge this yet. The results of the demo do not match those of the original cherab-solps demo. The plasma profiles should be the same, so I'm looking into what's causing this.

vsnever:
> @jacklovell, don't merge this yet. The results of the demo do not match those of the original cherab-solps demo. The plasma profiles should be the same, so I'm looking into what's causing this.

Everything works fine with the SOLPS plasma, so something is wrong with `get_edge_plasma()` or the json data.

vsnever:
Found it!
@skuba31, I'm sorry you have to fix our bugs, but could you please fix a copy-paste error in `get_edge_interpolators()`:
https://github.com/cherab/core/blob/0eb7454137b17bfbeca6d5e3eae4281f9cd4c157/cherab/generomak/plasma/plasma.py#L104
There should be "density" of course.
```python
    ne = Discrete2DMesh.instance(te, profiles["electron"]["density"], limit=False)
```
Also, I don't understand why the geometry transform is done like this
https://github.com/cherab/core/blob/0eb7454137b17bfbeca6d5e3eae4281f9cd4c157/cherab/generomak/plasma/plasma.py#L195
I think it should be
```python
    geometry_transform = translate(0, 0, z_range[0])
```
@Mateasek can you look at this?

skuba31:
@vsnever I have implemented the suggested changes. The results of the demo have changed, but I was not able to verify them, by running the original solps demo.

vsnever:
Yes, I'm happy with this PR. Great job @skuba31!

jacklovell:
Merged. Thanks for the great contribution @skuba31!

skuba31:
@jacklovell @vsnever Thank you for the valuable feedback.

Issue #211, opened 2020-01-31
Generomak - edge plasma specification


mattngc:
Following issue #209, we need an importable edge plasma specification. The example data could come from a representative 2D fluid code such as SOLPS-ITER or it could be completely hypothetical. The edge plasma profile should be consistent with the example equilibrium and core plasma developed in #210. It should also be anonymised (i.e. no particular Tokamak or source code). It should include:

- 2D triangular mesh extending from the plasma boundary (LCFS surface) all the way to first wall.
- electrons (temperature, density)
- main ions (temperature, density, flow velocities)
- representative impurities (C, Ne, etc)

vsnever:
Just as an option. 
I asked one of my colleagues, Andrey Pshenov, who is an experienced SOLPS user, whether SOLPS has test tokamak configuration as a part of its SDK, and it appears that SOLPS-ITER has some test simulation for the configuration similar to ASDEX. These data are used to test whether the program works correctly or not after compilation. I think that if the SOLPS-ITER team agree to share these test cases, you can get what you want in a simple way, namely without asking someone to create a custom SOL grid and perform the simulation specially for Cherab.
In case you decide to ask the SOLPS-ITER team about this, I think it’s Xavier Bonin whom you should contact.

Mateasek:
Being consistent with SOLPS examples is not a bad idea @vsnever , since it is far more difficult to setup and run.If I'm not mistaken, the mesh is defined not only by plasma facing components, but also by the equilibrium. We would have to adopt the equilibrium as the generomak case, wouldn't we?

vsnever:
> If I'm not mistaken, the mesh is defined not only by plasma facing components, but also by the equilibrium. We would have to adopt the equilibrium as the generomak case, wouldn't we?

Yes, but they may have equilibrium data for their test cases as well. As for the plasma facing components, the exact ASDEX first wall cannot be made public but any wall model that fits the 2D wall outline used in SOLPS simulations should be fine.

Mateasek:
I think this would be a good solution, taking what is there for one of the more difficult parts. It should be decided soon, because it has direct consequences for the other parts as PFCs and core plasma. Also the PFCs don't have to copy the simulation domain exactly I guess.

mattngc:
[...]

Issue #210, opened 2020-01-31
Generomak - core plasma specification


mattngc:
Following issue #209, we need an importable core plasma specification. The example data could come from a core 1D transport solver, such as METTIS or TRANSP. The profiles included should make sense for the example equilibrium and be anonymised (i.e. no particular Tokamak or source code). 1D profiles to include:

- electrons (temperature, density)
- main ions (temperature, density, rotation)
- representative profiles for an example impurity (C, Ne, etc)

The TMC proposed @Mateasek for this task and @jacklovell as the reviewer.

Mateasek:
Is the core and edge supposed to meet on the lcfs? 1D transport codes tend not to connect well with 2D edge simulations along the whole lcfs.

For now I suggest I will inspire myself with some JET profiles and create the model. This will be the easiest and most time efficient for now. Then we can rethink the approach and redo the core model once we have some edge model.

mattngc:
Lets see what happens in #211. Ideally, yes, they should line up... a little bit. It is after all a fake plasma, so you can always do some numerical editing of the profile at the boundary to get them in approximate agreement. The SOLPS guys might even have an AUG shot its based off, in which case you could just look up the shot and be "inspired" by the data.

CnlPepper:
> Is the core and edge supposed to meet on the lcfs? 1D transport codes tend not to connect well with 2D edge simulations along the whole lcfs.

It should matter too much. It doesn't have to be fully consistent as it's just for a demo of the framework. We've blended edge simulations with 1D core simulations using the Blend2D functions to crudely merge the boundary in the past.

Mateasek:
I think this can be closed.

PR #334, opened 2021-11-27
Add Generomak equilibrium

Mateasek:
This PR adds Generomak Equilibrium

The original equilibrium data were provided by @vsnever. 

The equilibrium includes the first wall contoure, which was used to create the first wall meshes. The contour is copying closely the solps simulation provided by @vsnever (at least in the important regions)
The psi grid was clipped to extend only over the vessel.
Positions of the magnetic axis and x-point were found for a denser (interpolated) grid (to reach a better precision)
The position of the x-point determined the new psi at the last closed flux surface (either I misunderstood what the original data meant or the value was completely off)
Last closed flux surface polygon found as a contour given the psi at x-point
The strike-points were found as an intersection of divertor legs contour and the first wall polygon

This PR is connected to issues #211 and #210


Mateasek:
I forgot to add that the f and q profiles were coppied from the example equilibrium because they were not in the equilibrium dataset.

vsnever:
Thanks you very much for this PR, @Mateasek. I think that `generomak_equilibrium.json` needs indents for better readability:
```python
with open(generomak_equilibrium.json', 'r') as f:
    eq=json.load(f)
with open(generomak_equilibrium.json', 'w') as f:
    json.dump(eq, f, indent=4)
```

Mateasek:
Yes, thanks for this. I will resubmit the PR, with the improved indentation, because it changes almost all the added lines.

PR #335, opened 2021-11-27
Add Generomak Equilibrium

Mateasek:
This PR adds Generomak Equilibrium

The original equilibrium data were provided by @vsnever.

The equilibrium includes the first wall contoure, which was used to create the first wall meshes. The contour is copying closely the solps simulation provided by @vsnever (at least in the important regions)
The psi grid was clipped to extend only over the vessel.
Positions of the magnetic axis and x-point were found for a denser (interpolated) grid (to reach a better precision)
The position of the x-point determined the new psi at the last closed flux surface (either I misunderstood what the original data meant or the value was completely off)
Last closed flux surface polygon found as a contour given the psi at x-point
The strike-points were found as an intersection of divertor legs contour and the first wall polygon

This PR is connected to issues #211 and #210


cherab/generomak/machine/data/equilibrium/generomak_equilibrium.json
```diff
@@ -0,0 +1,4630 @@
...
+            -0.776043274,
+            -0.8422613529999999,
+            -0.889208679,
+            -0.9450545650000001
+        ]
+    ],
+    "b_vacuum_radius": 1.48,
+    "b_vacuum_magnitude": 2,
+    "time": 0.0,
+    "f_profile": [
```
vsnever, line 215: This profile is inconsistent with the B<sub>vac0</sub> value. First of all, it's negative, while B<sub>vac0</sub> is positive. Also, at ψ<sub>norm</sub> = 1, it should give B<sub>vac0</sub> * R<sub>0</sub> = 2.96, but gives -2.93 instead. Can you please renormalize it?

Mateasek: Thanks for spotting this. I corrected the coordinate system and recalculated f and q profiles.

vsnever: Thanks, @Mateasek. Everything seems fine now.

vsnever (approved):

PR #336, opened 2021-11-29
Add Generomak edge plasma and the tools to create a plasma from 2D profiles.

### LINKED ISSUE #211: Generomak - edge plasma specification

vsnever:
This is a draft PR to fix #211. Description will be added soon.


Mateasek:
Yesterday I put together a set of json files which store the profiles. It might be a more suitable way how to store the data in git. I will push them in a separate PR, so you can look at it.

Mateasek:
@vsnever  please check PR #337

vsnever:
With velocities and magnetic field included, the json files will be ~4 times large than the pkl and probably much slower to read. However, this may not be a problem, let's ask @jacklovell's opinion on this.

Anyway, I think that the tools for creating plasma from the profiles defined on the triangular mesh and the Generomak edge plasma specifications must be separated. I'll close this PR and create a new one with the tools only.

Mateasek:
Why should it include also magnetic field? Isn't that included in the equilibrium?

The advantage of json is that it is well defined and stable format, that is not the case of a pickled class. Long term, this could be advantageous.

We definitely need tools loading the profiles to be separate. I will look into the core profiles once we get the equilibrium PR accepted. I will also look at how to effectively blend the 2D SOLPS profiles with the mapped-1D core profiles. There are many core transport codes which provide only 1D data and blending them properly with 2D edge results might be handy also outside the generomak profiles.

vsnever:
> Why should it include also magnetic field? Isn't that included in the equilibrium?

Yes, you are correct, it should not. For Generomak we should use the magnetic field from the equilibrium. I was still thinking in the frame of Cherab-SOLPS, sorry.

Nevertheless, `plasma_from_2d_profiles()` should include `b_field` optional parameter once `Discrete2DMesh` for `Vector3D` is added. The user may have only the profiles without the equilibrium.

Mateasek:
I'm sorry, but I'm still probably missing something. I thought that we want to add the profiles to have the possibility to build the plasma for Genormak, which would be used in examples. If this is the case, Generomak will have the equilibrium, which is consistent with the SOLPS simulation. I don't think providing an optional b-field  should be an option. If users want to add their own b-field, they can load the profiles separately and build a different plasma themselves, right?

vsnever:
> I'm sorry, but I'm still probably missing something. I thought that we want to add the profiles to have the possibility to build the plasma for Genormak, which would be used in examples. If this is the case, Generomak will have the equilibrium, which is consistent with the SOLPS simulation. I don't think providing an optional b-field should be an option. If users want to add their own b-field, they can load the profiles separately and build a different plasma themselves, right?

I was talking about the `plasma_from_2d_profiles()` function, which I propose to add to `tools/plasmas/mesh2d.py`. This is a general function, which allows to build a plasma from a set of profiles defined on a triangular mesh. It can be used outside the Generomak case.

jacklovell:
A factor of 4 size increase would mean ~10 MB of JSON, right? I think that's fine.

Agreed that the routines for creating the generomak profiles should be separate from the tools for loading it into plasma objects. I'm not sure whether a utility function is the best tool for making a plasma from mesh-defined quantities: perhaps a demo script would be better. Users may have data defined on a grid rather than a mesh, from which they build interpolators themselves. They may also have only partial data they want/need for their simulation, whereas this function will force them to specify everything.

vsnever:
I'm closing this because in #338 we decided that special tools for mesh plasma are not required and the work for Generomak edge plasma will continue in #337.

PR #337, opened 2021-11-29
Add Generomak sol plasma profile datasets

Mateasek:
The profiles are based on the solps example from Cherab-SOLPS module

The file mesh contains the vertex coords and triangle idicies. The rest of the files contain the species density and temperature profiles. The rotation profiles are for now skipped, but this is not a problem since most of the demonstrations can work without rotation.


Mateasek:
I forgot to add that the datasets were tested with the interpolators and it works. it Might be a good idea to combine it with contents of PR #336 o @vsnever

vsnever:
If we decide to store the edge plasma profiles in the set of json files, I would rename the `sol` folder to `edge` because scrape-off layer is a very special term. Also, we can have data on the magnetic field and velocities in json already now, even if we cannot use them yet. But let's wait for @jacklovell opinion first.

Mateasek:
I will rename the folder and I will add the velocity information.

Mateasek:
jfyi the added profiles now have 1.6MB, this will double with adding the velocity information. I would say that it is reasonable.

vsnever:
Thanks, @Mateasek. Currently, the SOLPS run for Generomak has velocities only for the ions. Andrey fixed the issue in SOLPS 4.3 and can now print fluxes of neutral particles, but given the amount of work he is currently doing, I don't want to bother him with this simulation again, at least for now.

vsnever:
I think that along with the `load_edge_plasma()`, which returns a Plasma object, we should have `load_edge_plasma_profiles()` which returns a dictionary of Generomak edge plasma profiles with `vertex_coords`, `triangles`, `electron_density`, `electron_temperature`, `species_density` and `species_temperature` keys. The values of `species_density` and `species_temperature` are dictionaries themselves with (element/isotope, charge) keys. Converting the data from this dict to `Discrete2DMesh` is trivial and can be performed in the demos when needed.

cherab/generomak/machine/plasma.py
```diff
@@ -0,0 +1,56 @@
+import os
```
jacklovell, line 1: Can you add the license header to this file please? Take it from any of the other `.py` files, but please make sure the dates are current.

jacklovell (changes requested):
Looks good to me. Only one minor change then I'm happy to merge.

vsnever:
I think this PR is missing `load_edge_plasma()` function, which would return a `Plasma` object constructed from edge profiles. Later, this function can be complimented with `load_core_plasma()` for core and `load_plasma()` for blended core and edge profiles as @Mateasek proposed in #338.

cherab/generomak/machine/plasma.py
Mateasek: Added licence which I coppied from a recently added file

Mateasek:
> I think this PR is missing `load_edge_plasma()` function, which would return a `Plasma` object constructed from edge profiles. Later, this function can be complimented with `load_core_plasma()` for core and `load_plasma()` for blended core and edge profiles as @Mateasek proposed in #338.

Added functions returning all the building blocks up to the plasma object.

cherab/generomak/machine/plasma.py
```diff
@@ -0,0 +1,213 @@
...
+
+                # get element or isotope
+                try:
+                    element = lookup_isotope(element_name)
+                except ValueError:
+                    element = lookup_element(element_name)
+                
+                # check charge consistency
+                if charge > element.atomic_number:
+                    raise ValueError("Charge {:d} exceeds the maximum possible charge for the element {}.".format(charge, element.name))
```
vsnever, line 92: You are already know the element name and charge from the loops. I would just check the data sanity instead of this.
```python
if element.name != element_name:
   raise RuntimeError('Inconsistent data in {}: expected {}, got {}.'.format(path, element.name, element_name))
if chrg != charge:
   raise RuntimeError('Inconsistent data in {}: expected {}, got {}.'.format(path, chrg, charge))
```

cherab/generomak/machine/plasma.py
```diff
@@ -0,0 +1,213 @@
...
+    vertex_coords = np.asarray(mesh["vertex_coords"])
+    r_range = (vertex_coords[:, 0].min(), vertex_coords[:, 0].max())
+    z_range = (vertex_coords[:, 1].min(), vertex_coords[:, 1].max())
+    plasma_height = z_range[1] - z_range[0]
+
+    padding = 1e-3 #enlarge plasma geometry for safety
+
+    outer_column = Cylinder(radius=r_range[1] + padding, height=plasma_height + 2 * padding)
+    inner_column = Cylinder(radius=r_range[0] - padding, height=plasma_height + 4 * padding)
+    inner_column.transform = translate(0, 0, -2 * padding)
```
vsnever, line 190: Can you please clarify how adding this padding improves the safety? In `RayTransferBox` and `RayTransferCylinder`, the geometry of the bounding primitive needed to be reduced a little to avoid out-of-range errors for cell indices. But what could go wrong here?

cherab/generomak/machine/plasma.py
```diff
@@ -0,0 +1,213 @@
...
+        for stage, stage_data in elem_data.items():
+            species = Species(stage_data["element"], stage, stage_data["distribution"])
+            plasma_composition.append(species)
+
+    # Populate plasma
+    plasma = Plasma()
+    plasma.geometry = plasma_geometry
+    plasma.atomic_data = atomic_data
+    plasma.electron_distribution = dists["electron"]
+    plasma.composition = plasma_composition
```
vsnever, line 210: Since we already have equilibrium, is it possible to add `b_field` from there?

vsnever (changes requested):
Great, thank you! 
I have a few comments though.

cherab/generomak/machine/plasma.py
Mateasek: The interpolators are set to give 0 outside the triangles, which should make it save to extend the bounding primitive. I wanted to make sure that the bounds include the whole mesh, some SOLPS triangles in the divertor region can be very small, but you are right that not using padding would not influence the results, especially the demos. Should I remove it?

vsnever: The bounds `r_range` and `z_range` already include the whole mesh no matter how small the triangles are. I would remove this code because it can be confusing, especially in adding different number of padding to the different parameters of the primitives. But maybe I'm missing something, so let's ask for @jacklovell's opinion.

cherab/generomak/machine/plasma.py
Mateasek: I'm not sure I understand what you mean.

I realised the check of the species description is premature here. The data and interpolators should be returned even for a description which is not matching Cherab list of elements and isotopes (or violates the charge constrains). I will not pass the element. Also, the charge check is done when the Species object is created, so it is redundant here.

cherab/generomak/machine/plasma.py
jacklovell: If the padding is unnecessary I would remove it, and keep the function as simple as possible.

cherab/generomak/machine/plasma.py
vsnever: I just didn't get why you had to create the element from the element name while you already have it as a loop variable. The same was for the charge.

I think, I understand the updated code more. So, the element and charge loops are used only to avoid listing through all the json files in the folder and separating those for the species from the others by accessing the respective keys. That's fair since the file names are fixed.

Mateasek: ah, right. Well the element data are read from files and I thought it should be validated, but I think removing int is a good choice (DRY). I only used the element instances to get the numbers and names (not because I would not remember them).

cherab/generomak/machine/plasma.py
Mateasek: Removed padding to enlarge the plasma primitive. Padding stayed in the case of inner column to avoid subtraction of parallel plains

cherab/generomak/machine/plasma.py
Mateasek: Good point. Added b_field

vsnever: Thanks, the magnetic field will be needed for high-resolution spectroscopy demos.

Mateasek:
> Sorry for noticing this late, but the directory tree does not match that of the Cherab-JET subpackage, https://github.com/cherab/jet/tree/development/cherab/jet. I think the machine folder should contain only the hardware components, like the first wall, while equilibrium and plasma should be in the separate sub-folders of cherab/generomak.

You are right, good point. I will move the python files, the data will stay. I will also create a separate PR, where I will move the equilibrium.py.

> Also, I think it's worth adding parent and name arguments to get_edge_plasma() with the default name Generomak edge plasma.

Added the optional parameters

vsnever:
Thanks. I think the data should be moved as well, namely to `generomak/plasma/data/edge`. And the same for the equilibrium data, `generomak/equilibrium/data`.

Mateasek:
Here I disagree. I think it is better to concentrate data in a single directory instead of having 3 separate data directories.

Mateasek:
I just notticed the thumbs up, it seems that both @vsnever and @jacklovell agree with separating it, so I will move it.

jacklovell:
By way of precendent, `cherab.tools.equilibrium.example_equilibrium` has its data kept in the equilibrium directory. So keeping the data with the object of interest would be consistent with this, rather than having a separate data store for everything. Certainly I don't think it makes sense to have the plasma data in the `machine` directory.

vsnever:
Looks like #341 must be merged first.

Mateasek:
> Looks like #341 must be merged first.

yes, the equilibrium is used to get the magnetic field.

Mateasek:
Seems like this is good to go @jacklovell

jacklovell (approved):

PR #345, opened 2021-12-07
[fix] correct the number of args.

munechika-koyo:
Because of missing Intersection args, the compile error has occurred.
compile error message:

>  cherab/tools/raytransfer/roughconductor.pyx:35:10: Call with wrong number of arguments (expected 11, got 10)

### DIFF
--- a/cherab/tools/raytransfer/roughconductor.pyx
+++ b/cherab/tools/raytransfer/roughconductor.pyx
@@ -18,7 +18,7 @@
 # under the Licence.
 #
 
-from raysect.optical cimport Point3D, Vector3D, AffineMatrix3D, World, Ray, Spectrum, new_vector3d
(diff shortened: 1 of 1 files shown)


jacklovell:
This change was introduced into the development branch of raysect, which will become version 0.8.x but is not yet released. Going forwards we're going to keep the Cherab development branch compatible only with the most recent released version of raysect (see #320 for a discussion of this). When I get time I'll merge this into a new `raysect-0.8-compat` branch so if you or others absolutely need unreleased raysect features then this new branch will be suitable. But this'll need to wait until after the imminent Cherab 1.3.0 release.

jacklovell:
Note also that there are multiple places in the code where the `Intersection` argument will need to be added.

munechika-koyo:
I also thought what you taught me was right because it did not pass all of the Cl /Run tests.
Sorry for my sudden PR and I'll help address this compatible problem if you want.

PR #347, opened 2021-12-07
Improve the docstrings in OpenADAS class.

vsnever:
This PR makes the docstrings in `OpenADAS` class more helpful, in particular adding information about the interpolation and extrapolation methods used for each rate coefficient. The need to add this information was discussed in #304.


jacklovell (approved):
Looks fine to me. Thanks.

PR #348, opened 2021-12-10
make EFITquilibrium class variables public

### LINKED ISSUE #349: Make EFITequilibrium class variables public
As mentioned in PR https://github.com/cherab/core/pull/348, I threw the issue about it.
Why I proposed this issue is that I need to reconstruct some function classes (e.g. `MagneticField`) and apply them to `EFITequilibrium` variables to use my own tokamak device and calculated equilibrium information.

munechika-koyo:
To inherit EFITquilibrium class and override some members, I change the access right them into public. However member variables must be private in usual, so I think it might be preferable to define them as properties.

Anyway, I just throw PR, and if this is Inappropriate, I will close it immediately.

### DIFF
--- a/cherab/tools/equilibrium/efit.pxd
+++ b/cherab/tools/equilibrium/efit.pxd
@@ -32,18 +32,18 @@ cdef class EFITEquilibrium:
 
 
     cdef:
-        readonly Function2D psi, psi_normalised
-        readonly double psi_axis, psi_lcfs
-        readonly tuple r_range, z_range
-        readonly Point2D magnetic_axis
-        readonly tuple x_points, strike_points
-        readonly VectorFunction2D b_field, toroidal_vector, poloidal_vector, surface_normal
-        readonly Function2D inside_lcfs, inside_limiter
-        readonly Function1D psin_to_r
-        readonly double time
-        readonly np.ndarray lcfs_polygon, limiter_polygon
-        readonly np.ndarray psi_data, r_data, z_data
-        readonly Function1D q
+        public Function2D psi, psi_normalised
+        public double psi_axis, psi_lcfs
+        public tuple r_range, z_range
+        public Point2D magnetic_axis
+        public tuple x_points, strike_points
+        public VectorFunction2D b_field, toroidal_vector, poloidal_vector, surface_normal
+        public Function2D inside_lcfs, inside_limiter
+        public Function1D psin_to_r
+        public double time
+        public np.ndarray lcfs_polygon, limiter_polygon
+        public np.ndarray psi_data, r_data, z_data
+        public Function1D q
         double _b_vacuum_magnitude, _b_vacuum_radius
         Function1D _f_profile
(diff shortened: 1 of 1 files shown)


vsnever:
Hi @munechika-koyo,
It is assumed that the user can set public attributes without worrying to break anything. In the case of `EFITEquilibrium`, setting most of the attributes after initialization will result in data inconsistencies, that's why they were protected with read only access (not a good protection for arrays, as the user can still change the elements, but this issue affects many Cherab classes). So, I think the best way to solve the inheritance problem is to define the attributes as properties. Most, if not all, of these properties will not need setters. However, let's wait for other opinions.

Mateasek:
There was a plan to completely rework the equilibrium class, with more levels of detail (2D, 3D). I always viewed that what we have is a preliminary state which needs a lot of work. @vsnever  is right that the attributes would be better. I would anyway prefer if @munechika-koyo could open an issue connected to this PR so we can talk about it there.

munechika-koyo:
I threw the issue https://github.com/cherab/core/issues/349
I would appreciate it if someone would link this PR to the issue.

CnlPepper:
> (not a good protection for arrays, as the user can still change the elements, but this issue affects many Cherab classes).

In python spcae, any array returned by an object, where changes to elements outside the object could cause inconsistency (e.g. stale caching), should be returned as a copy, not a reference.

In cython, for speed, we allow direct access. However the user is expected to obey any documentation warning of data inconsistency if elements are changed. Given cython users will tend to be more experienced engineers, we can expect a higher degree of understanding of the risks.

vsnever:
> In python spcae, any array returned by an object, where changes to elements outside the object could cause inconsistency (e.g. stale caching), should be returned as a copy, not a reference.

Thanks, @CnlPepper. In case when the arrays are not meant to change even within the object, like in the case of `EFITEquilibrium`, there is also an option to set the `writeable` flag to False. So, there are two different approaches in this case:

Assign a copy, return a copy:
```python
@property
def some_array_property(self):
    return np.copy(self._some_array_property)

@some_array_property.setter
def some_array_property(self, value):
    self._do_some_checks(value)
    self._some_array_property = np.copy(value)
    self._update_dependent_attributes()
```
Assign a copy, return a read-only reference:
```python
@property
def some_array_property(self):
    return self._some_array_property

@some_array_property.setter
def some_array_property(self, value):
    self._do_some_checks(value)
    self._some_array_property = np.copy(value)
    self._some_array_property.flags.writeable=False
    self._update_dependent_attributes()
```
What's the best practice, or is it a case-by-case basis?

jacklovell:
I'm going to close this. Based on the discussion in #349, the best solution is to re-factor the equilibrium tools into a base Equilibrium object which can then be subclassed to store equilibria from different tools (EFIT, TSC etc.). The existing `EFITEquilibrium` class is intended to be a static representation of the result of an EFIT reconstruction only.

PR #352, opened 2022-01-24
Add more accurate Gaunt factor for Bremsstrahlung emission model

### LINKED ISSUE #351: Improve Gaunt factor in Bremsstrahlung emission model
This issue was briefly discussed at ITER Synthetic Diagnostic meeting, so I'm putting it here.

The applicability of the Gaunt factor, _G_, in Bremsstrahlung emission model, 
https://github.com/cherab/core/blob/0eb7454137b17bfbeca6d5e3eae4281f9cd4c157/cherab/core/model/plasma/bremsstrahlung.pyx#L99-L100
is very limited. It's missing the dependance on the wavelength and Zeff, which means that it can only be used within a limited range of the visible spectrum and only for the hot plasma, Te >~ 100 eV.

This Gaunt factor is compared with classical and Born approximations for the Gaunt factor in this ITER report, [3338YT](https://user.iter.org/default.aspx?uid=3338YT).

Briefly, _G<sub>class</sub>_ can be used if _α ≡ Zeff/(Te/Rd)<sup>1/2</sup>_ ≫ 1 and _G<sub>Born</sub>_ can be used if _α_ ≪ 1. Here:
_G<sub>class</sub>_ ≈ √3/π ln(2.e-4 · Te<sup>3/2</sup>[eV]/Zeff · λ[nm] + exp(π/√3))
and
_G<sub>Born</sub>_ ≈ 0.55 · ln(0.0018 λ[nm] Te[eV]) (low frequency limit).

The easiest solution would be to use a combination of _G<sub>class</sub>_ and _G<sub>Born</sub>_ in Cherab. The following two figures show _G<sub>class</sub>(Te, λ)_ and _G<sub>Born</sub>(Te, λ)_ as well as their difference (left plot) and combination, _G<sub>eff</sub>_, for Zeff = 1 and Zeff = 2. The vertical line shows _α=1/√2_ 

Using _G<sub>class</sub>_ for Te < 2 Rd Zeff <sup>2</sup> and _min_(_G<sub>class</sub>_, _G<sub>Born</sub>_) for Te > 2 Rd Zeff <sup>2</sup> will improve accuracy for low temperatures and in infrared spectral range.
gaunt_factor_born_vs_classic_zeff_1 0
gaunt_factor_born_vs_classic_zeff_2 0

I would like to hear other opinions on this, so we probably continue discussions at ITER Synthetic Diagnostic meetings.

vsnever:
This fixes #351.

- The new classes: `FreeFreeGauntFactor`, `InterpolatedFreeFreeGauntFactor` and `MaxwellianFreeFreeGauntFactor` are added in `cherab.core.atomic`.
- The data for `MaxwellianFreeFreeGauntFactor` is taken from M.A. de Avillez and D. Breitschwerdt, "Temperature-averaged and total free-free Gaunt factors for κ and Maxwellian distributions of electrons", 2015, Astron. & Astrophys. 580, [A124](https://www.aanda.org/articles/aa/full_html/2015/08/aa26104-15/aa26104-15.html) (Table A.1).
- The Born approximation is used outside the interpolation range.
- A new method `free_free_gaunt_factor()` is added to the `AtomicData` interface.
- `Bremsstrahlung` emission model now takes the Gaunt factor from the `AtomicData` interface.
- An indexing error in `Bremsstrahlung` model is fixed.
- Bremsstrahlung demo is added.


cherab/core/atomic/gaunt.pyx
```diff
@@ -0,0 +1,152 @@
...
+        :param double temperature: Electron temperature in eV.
+        :param double wavelength: Spectral wavelength.
+
+        :return: free-free Gaunt factor
+        """
+        raise NotImplementedError("The evaluate() virtual method must be implemented.")
+
+    def __call__(self, double zeff, double temperature, double wavelength):
+        """
+        Returns self.evaluate(zeff, temperature, wavelength).
```
jacklovell, line 54: Just use the docstring of `evaluate` here

jacklovell (changes requested):
Looks good. Can you please add an entry to the changelog? I think this'll go into 1.4.0.

vsnever:
Thanks for the review @jacklovell, I've updated the changelog.

cherab/core/atomic/gaunt.pyx
vsnever: Done.

jacklovell (approved):

vsnever:
Just realised that this if statement in `Bremsstrahlung` model:
https://github.com/cherab/core/blob/0246b9d6dc037c02f5e2a4b9eb8b4e8558292eef/cherab/core/model/plasma/bremsstrahlung.pyx#L73-L75
does nothing, because `plasma.z_effective()` raises `ValueError` when Zeff is zero:
https://github.com/cherab/core/blob/b84f00973b930f81d096a183ef6aa683d7a666b8/cherab/core/plasma/node.pyx#L422-L431
So, we can use exception in `Bremsstrahlung`:
```python
try:
    z_effective = self._plasma.z_effective(point.x, point.y, point.z)
except ValueError:
    return spectrum
```
or we can return zero instead of raising `ValueError` in `plasma.z_effective()`. 

I think `z_effective()` should only throw an exception if the plasma contains no charged particles at all, not if their densities are zero at this particular point in space. @jacklovell, what do you think?

vsnever:
I realised that because the Gaunt factor now depends on Z, we cannot use Z<sub>eff</sub> anymore. Instead we have to sum over all charged species in the `Bremsstrahlung.emission()`. I'll try to do this now, but until it's done I'm converting this to draft.

vsnever:
The move towards summation over charged species in `Bremsstrahlung` model inevitably affects performance. I tried to minimize the effect by caching the species densities before the spectral loop, but for a plasma with a lot of species, the slowdown will be significant. However, I prefer accuracy over performance here. The equation with Z<sub>eff</sub> is used to interpret the Bremsstrahlung signal and reconstruct Z<sub>eff</sub>, if we use the same simplification to generate the synthetic signal, the user will lose the opportunity to test the accuracy of reconstruction.

Mateasek:
Hi @vsnever,

thanks for this PR, it was needed for a long time. 

Regarding the zero density of charged species and Zeff behaviour, why should the method raise an error instead of simply returning 0? We allow temperature to be 0 and we let the models cope with it. Shouldn't we adapt the same approach here? 

I have one question. You store the mandatory data in the cherab.core. Wouldn't it be better, to make a new package, something like cherab.default_data or sorts, where we would be storing similar cases? The atomic data were separated for a good reason and we can start providing a default set, as we do with OpenADAS.

The last question I have, shouldn't the gaunt factor become an optional parameter of the Bremsstrahlung emission model? This would allow more freedom in the future.

Mateasek (commented):
Hi @vsnever,

thanks for this PR, it was needed for a long time.

Regarding the zero density of charged species and Zeff behaviour, why should the method raise an error instead of simply returning 0? We allow temperature to be 0 and we let the models cope with it. Shouldn't we adapt the same approach here?

I have one question. You store the mandatory data in the cherab.core. Wouldn't it be better, to make a new package, something like cherab.default_data or sorts, where we would be storing similar cases? The atomic data were separated for a good reason and we can start providing a default set, as we do with OpenADAS.

The last question I have, shouldn't the gaunt factor become an optional parameter of the Bremsstrahlung emission model? This would allow more freedom in the future.

jacklovell:
>>   I have one question. You store the mandatory data in the cherab.core. Wouldn't it be better, to make a new package, something like cherab.default_data or sorts, where we would be storing similar cases? The atomic data were separated for a good reason and we can start providing a default set, as we do with OpenADAS.

> A good idea. If @jacklovell approves, I'll move the InterpolatedFreeFreeGauntFactor and MaxwellianFreeFreeGauntFactor as well as the json files to cherab.default_data.

Agreed, a separate data store sounds like a good idea.

vsnever:
> Agreed, a separate data store sounds like a good idea.

So, in addition to `AtomicData` from `cherab.core.atomic` we introduce `DefaultAtomicData(AtomicData)` in `cherab.default_data` and `OpenADAS` will inherit from `DefaultAtomicData`, is that correct?

Mateasek:
Should we create an issue abut the Default data? I don't think `OpenADAS` has to be inheriting from `DefaultAtomicData` or I'm missing something? The base classes should be in core and the radiation model can take any of the data packages, right?

vsnever:
> I don't think OpenADAS has to be inheriting from DefaultAtomicData or I'm missing something? The base classes should be in core and the radiation model can take any of the data packages, right?

As a user, I would like to have both bremsstrahlung data and atomic rates or photon emission coefficients in the same atomic data source. "DefaultAtomicData" has the Gaunt factor needed for bremsstrahlung, but lacks atomic rates. `OpenADAS`, if inherited from the core `AtomicData`, will have the rates but not the Gaunt factor. Therefore, it will not be possible to simulate, for example, spectral line emission on top of the bremsstrahlung background without initializing line emission models with `OpenADAS` and `Bremsstrahlung` model with `DefaultAtomicData`. But the most convenient way to use atomic data is by connecting it to the `Plasma` object. Thus, I think third party atomic datasources should extend and/or enhance the default atomic datasource. So, if some data is missing in the third party datasource, the default data is used. Otherwise, we risk confusing the user.

PR #353, opened 2022-01-28
Packaging improvements

jacklovell:
A few changes here to make installation of Cherab more robust and more accessible for end users. Recent improvements in Python packaging (PEP517, PEP518) mean we can strip out some of the workarounds which were historically required to install the package, and follow modern best practices.

* Update README and docs to recommend using pip rather than calling setup.py directly, as the latter is deprecated.
* Exploid PEP518 and PEP517 to specify Cython as a build dependency instead of a runtime dependency.
* Remove logic in setup.py to behave differently if Cython is not available.

PEP517/PEP518 support in pip is available as early as version 10.0 back in early 2018, so I think it's pretty safe to assume new installations of Cherab will be in environments where this is supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,56 +12,66 @@ for guidance on using the code.
 Installation
 ------------
 
-Cherab is a large code framework consisting of a core package and feature  
-packages. Users will generally install the core package and the specific  
-feature packages they need for their work. For example, users working on the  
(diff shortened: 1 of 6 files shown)


jacklovell:
Requesting @CnlPepper's thoughts on this. Some of the changes may be applicable to raysect too.

PR #357, opened 2022-02-22
Enable testing of SART OpenCL inversion routines with POCL

jacklovell:
Currently the tests for the OpenCL SART solvers are skipped because they require an OpenCL driver. This PR configures the tests to use the POCL CPU-only driver so we can run the tests through Github Actions.

The tests do output some debugging information from the solvers, which should really be put behind an `if verbose` flag or use a debug-level logger. May be worth fixing before merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,9 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Python dependencies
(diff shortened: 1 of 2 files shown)


vsnever (approved):
Thanks for this @jacklovell. I think this can be merged as it is. 

In a separate PR, I'll update the code for `SartOpencl` and the function in `opencl_utils.py` to hide all non-critical output if `verbose` is set to `False`, which will be the default. In the same PR I'll make all `SartOpencl` attributes protected as they are now public for no reason.

PR #365, opened 2022-03-15
Feature/toroidal mesh from polygon

wave46:
Added `toroidal_mesh_from_polygon()` for making mesh not fully-360 degrees axysimmetric elements and made `axisymmetric_mesh_from_polygon()` as its wrapper.

### DIFF
--- a/cherab/tools/primitives/axisymmetric_mesh.pyx
+++ b/cherab/tools/primitives/axisymmetric_mesh.pyx
@@ -23,15 +23,14 @@ from .toroidal_mesh import toroidal_mesh_from_polygon
 
 cpdef Mesh axisymmetric_mesh_from_polygon(object polygon, int num_toroidal_segments=500):
     """
-    .. deprecated:: 1.4.0
-       Use `toroidal_mesh_from_polygon` with `toroidal_extent=360` instead.
     Generates an Raysect Mesh primitive from the specified 2D polygon.
 
-    :param np.ndarray polygon: A numpy array with shape [N,2] specifying the wall outline polygon
-      in the R-Z plane. The polygon should not be closed, i.e. vertex i = 0 and i = N should not
-      be the same vertex, but neighbours.
+    :param object polygon: An object which can be converted to a numpy array with shape [N,2] 
+                           specifying the wall outline polygon in the R-Z plane. The polygon 
+                           should not be closed, i.e. vertex i = 0 and i = N should not be the
+                           same vertex, but neighbours.
     :param int num_toroidal_segments: The number of repeating toroidal segments that will be used
-      to construct the mesh.
+                                      to construct the mesh.
     :return: A Raysect Mesh primitive constructed from the R-Z polygon using symmetry.
 
     .. code-block:: pycon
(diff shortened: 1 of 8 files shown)


cherab/tools/primitives/axisymmetric_mesh.pyx
```diff
@@ -1,24 +1,37 @@
...
 
+from .toroidal_mesh import toroidal_mesh_from_polygon
 
-cdef double DEG2RAD = 2 * np.pi / 360
 
-
-cpdef Mesh axisymmetric_mesh_from_polygon(np.ndarray polygon, int num_toroidal_segments=500):
+cpdef Mesh axisymmetric_mesh_from_polygon(object polygon, int num_toroidal_segments=500):
     """
     Generates an Raysect Mesh primitive from the specified 2D polygon.
```
vsnever, line 26: The only reason to keep this function is for backwards compatibility. I think we can add a deprecation warning. Could you add the following text to the beginning of the docstring?
```
.. deprecated:: 1.4.0
       Use `toroidal_mesh_from_polygon` with `toroidal_extent=360` instead.
```

vsnever (changes requested):
Thanks for this contribution, @wave46. I think it's good to have some compatibility with ToFu meshes.

Can you add toroidal_mesh_from_polygon to `docs/source/tools/primitives.rst` and update the `CHANGELOG.md`?

jacklovell (changes requested):
Thanks for this @wave46. I have a couple of minor comments on the code which should be simple to fix, and a more general conceptual comment.

We actually need to be a little bit careful about the `toroidal_extent=360` case, as it does produce a different shape (no end faces). It also I think is likely to serve a different purpose: its use case in the past has been to easily construct a 3D first wall from a 2D limiting surface (from EFIT, originally, but also for geometry from 2D modelling codes such as SOLPS) as a quick and dirty approximation in the absence of proper 3D CAD. In contrast, the new `toroidal_mesh_from_polygon` appears to be more geared towards toroidal extrusions of smaller parts of the PFCs and is clearly making more of an effort at fidelity in the absence of CAD than a simple axisymmetric approximation. From a code-reuse perspective having `axisymmetric_mesh_from_polygon` be a special case of `toroidal_mesh_from_polygon` is a good thing, but I worry it'll be less obvious for end users if we deprecate/remove `axisymmetric_mesh_from_polygon` entirely.

To summaries this rambling, I think what I'm getting at is "a function should do one thing and do it well". My question is whether making an axisymmetric mesh can be considered sufficiently different from making a toroidal extrusion to deserve to keep its own function, or whether as @vsnever suggests they're ultimately the same thing from an end-user's perpsective.

vsnever:
I agree with @jacklovell, the fact that `axisymmetric_mesh_from_polygon` is now a wrapper, does not mean it should be deprecated. Let's keep it, but in this case `toroidal_extent` indeed should be a positional argument with no default value.

Also, @wave46, could you please add a check for `toroidal_extent` range:  `0 < toroidal_extent <= 360`?

jacklovell (approved):

jacklovell:
Thanks for the contribution @wave46

Issue #372, opened 2022-08-04
Add blended profiles for Generomak plasma


vsnever:
As already discussed in #360, a function that returns full Generomak plasma, which includes both the core and edge regions, is needed. Since the edge profiles also cover the thin outer region of the core, the mask used to blend the core and edge profiles can be a function of ψ<sub>norm</sub>.

jacklovell:
Fixed by #373

PR #373, opened 2022-08-04
Add full profiles for Generomak plasma

vsnever:
This PR solves #372 and contains the following changes in `generomak/plasma/plasma.py`:

- Adds the function `get_full_profiles()`, which blends core and edge profiles to form the full 2D profiles using the linear mask function, F(ψ<sub>norm</sub>), as a modulator.
- Adds the function `get_plasma()`, which creates Generomak plasma from specified equilibrium and distributions. By default this function returns the plasma built from the full plasma profiles.
- Simplifies `get_core_plasma()` and `get_edge_plasma()` as both functions now call the new `get_plasma()` function.
- Removes the `distributions` argument from `get_core_plasma()`, because for non-default distributions the general `get_plasma()` function can be used directly.
- Corrects the default LCFS densities for neutral species and low ionised carbon in `get_core_profiles_description()` to better match the edge profiles.
- Fixes the toroidal rotation profile in `get_core_profiles_description()`. @Mateasek, could you please check if this fix is correct?
- Renames the function `get_edge_distributions()` to `get_2d_distributions()` and adds an optional `profiles_2d` argument, since this function can now be used with both edge and full 2D profiles. By default, the function returns distributions built with edge profiles.
- Slightly changes the structure of the dictionary returned by `get_2d_distributions()` to match the structure of the dictionary returned by `get_core_distributions()`. Since both dictionaries now have the same structure, they can be used with the `get_plasma()` function.
- Replaces all zero velocities in default profiles with negligible but non-zero values to avoid errors when blending.

Some outliers in the temperature profiles in `hydrogen0.json` and `carbon0.json` are replaced with reasonable values.

Two demos are added, both plotting the core, edge and blended Generomak 2D plasma profiles. The first demo uses the interpolators directly, while the second one samples the profiles from the `Plasma` objects.


vsnever:
The `Raysect-0.8-compat` branch should be merged into the `development` branch to pass the tests.

jacklovell:
Thanks for this Vlad. I'm a bit snowed under at the moment, but will take a look at this (and also merging raysect-0.8 into development) when I get time in a couple of weeks.

Mateasek:
Hi @vsnever,

thanks for the review, I will find some time next week to look at this.

cherab/generomak/plasma/plasma.py
```diff
@@ -419,15 +390,23 @@ def get_core_profiles_description(lcfs_values=None, core_args=None):
...
         lcfs_values = get_edge_profile_values(r, z)
 
+        # manually correcting the LCFS densities for neutral species and low ionised carbon
+        # to better match the edge profiles
+        lcfs_values["composition"]["hydrogen"][0]["density"] = 1.e15
+        lcfs_values["composition"]["carbon"][0]["density"] = 1.e5
+        lcfs_values["composition"]["carbon"][1]["density"] = 1.5e5
+        lcfs_values["composition"]["carbon"][2]["density"] = 1.e9
+        lcfs_values["composition"]["carbon"][3]["density"] = 1.e13
+
```
jacklovell, line 400: Can we correct these values at source, rather than hidden away here?

vsnever: Just realized that there is a more serious problem here, the core profiles do not match plasma neutrality. We need to use the ionisation balance solver. I'll try to fix this.

vsnever:
With the last commit, I fixed the plasma neutrality issue for core profiles. The charge state distributions are now calculated using the ionisation balance tools. Also, manual adjustment of the LCFS values is no longer required, the core profiles match the profiles in the edge plasma quite well. However, the core profiles now depend on the atomic data that we have no control over. I think, this is not a big issue, because almost all Cherab demos use atomic data anyway.

jacklovell:
> However, the core profiles now depend on the atomic data that we have no control over. I think, this is not a big issue, because almost all Cherab demos use atomic data anyway.

You could avoid this by pre-computing the profiles using ionisation balance but then storing the computed values rather than computing them at instantiation time. The data then effectively depends on the atomic data at generation time but end users don't need to have it available when using Generomak.

vsnever:
> You could avoid this by pre-computing the profiles using ionisation balance but then storing the computed values rather than computing them at instantiation time. The data then effectively depends on the atomic data at generation time but end users don't need to have it available when using Generomak.

I added json files with precalculated core profiles as well as two functions to handle them: `load_core_profiles()` and `get_core_interpolators()`. All other functions and demo scripts now use `get_core_interpolators()` instead of `get_core_profiles_description()` by default.

The density profiles are very sharp functions near Ψ<sub>norm</sub>=1, so to take this into account I changed how Ψ<sub>norm</sub> is sampled for ionisation balance solver:
`psin_1d = np.append(1. - np.geomspace(1.e-4, 1, 1023)[::-1], [1.])`.

In json files Ψ<sub>norm</sub> is sampled in the same way:
`psin_norm = np.append(1. - np.geomspace(1.e-4, 1, 255)[::-1], [1.])`.

PR #375, opened 2022-08-18
Fix typo in Bremsstrahlung docstrings

skuba31:
Comparing to the equation in [referenced article](https://doi.org/10.1063/1.2969097), there is a missing minus in the exponent of electron temperature.


vsnever (approved):
Thanks, @skuba31. I'll fix this in #352 as well.

PR #376, opened 2022-08-18
Add quiet property to Observer0DGroup

skuba31:
The quiet property was unintentionally left out from the initial observer groups implementation (#332). This PR adds support for it.


vsnever (changes requested):
Hi @skuba31, thanks!

I just discovered that another property that has been overlooked is `ray_importance_sampling`. Could you add it and also update the docstring?

I think both properties were overlooked first by me in #284 and then by you in #332, because in `_ObserverBase` they are not really properties, but attributes, and therefore have no setters.

skuba31:
Hi Vlad

I have reordered the attributes in Group to match _ObserverBase and Observer0D to make sure I am not missing anything else. I have added `ray_importance sampling` and `spectral_rays`

cherab/tools/observers/group/base.py
```diff
@@ -32,22 +32,22 @@ class Observer0DGroup(Node):
...
 
     :ivar list names: A list of observer names.
-    :ivar list pipelines: A list of all pipelines connected to each observer in the group.
     :ivar list/RenderEngine render_engine: Rendering engine used by the observers.
                                            Note that if the engine is shared, changing its
                                            parameters for one observer in a group will affect
                                            all observers.
-    :ivar list/float min_wavelength: Lower wavelength bound for sampled spectral range.
+    :ivar list/int spectral_bins: The number of smaller sub-spectrum rays the full spectrum will be divided into.
+    :ivar list/int spectral_rays: The number of smaller sub-spectrum rays the full spectrum will be divided into.
```
vsnever, line 40: Thank you! Everything is fine except for this copy-paste error in the docstring.

skuba31: It should be fixed now.

vsnever (approved):

jacklovell:
@skuba31 Thanks for this. Could you please rebase onto current development so we can check all the tests pass? I'm happy to merge it after that.

Issue #364, opened 2022-03-08
Default atomic data distributed with Cherab


vsnever:
While discussing where to store the Gaunt factor for `Bremsstrahlung` emission model in #352, @Mateasek proposed to create a subpackage `cherab.default_data` to store the default atomic data distributed with Cherab.

This was agreed, but the question remained whether the third party atomic data sources, like `OpenADAS`, should inherit from the core `AtomicData` interface or from the `DefaultAtomicData` interface.

I think that the third party data sources should inherit from the `DefaultAtomicData`, because the `DefaultAtomicData` may contain the data not present in the third party data source. For example, `DefaultAtomicData` has the Gaunt factor needed for bremsstrahlung, but lacks atomic rates, while `OpenADAS`, if inherited from the core `AtomicData`, will have the rates but not the Gaunt factor. Therefore, it will not be possible to simulate, for example, spectral line emission on top of the bremsstrahlung background without initializing line emission models with `OpenADAS` and `Bremsstrahlung` model with `DefaultAtomicData`. But the most convenient way to use atomic data is by connecting it to the `Plasma` object and thus using the same data source for all models.

We may run into a case when both the `DefaultAtomicData` and the `OpenADAS` contain data for the same physical quantity. For such a case  we may add a parameter `override_default` to `OpenADAS`, which, if `True`, will use the `OpenADAS` data, and if `False`, the default data.

@Mateasek, @jacklovell, what do you think?

Mateasek:
Thanks for opening an issue @vsnever. I think that creating the default_data could help in the future. It is in my opinion much better than storing any data in the core.

Another option is importing classes/functions from the default_data in the "derived" data. Combining both imports and inheritance could give us enough flexibility in the future. The main point I see is to give users a flexible and simple way how to build their own Cherab data source. Without the need of copying the actual data between repositories. For example, in future, we can also have ALADDIN data source and users could combine data from both OpenADAS, ALADDIN and default_data to form their own repository.

Here I think I would also ask for opinion of @CnlPepper and @mattngc, because this is an important decision to take.

vsnever:
> Another option is importing classes/functions from the default_data in the "derived" data. Combining both imports and inheritance could give us enough flexibility in the future. The main point I see is to give users a flexible and simple way how to build their own Cherab data source. Without the need of copying the actual data between repositories. For example, in future, we can also have ALADDIN data source and users could combine data from both OpenADAS, ALADDIN and default_data to form their own repository.

[...]

PR #377, opened 2022-08-25
Split the openadas module into atomic repository and OpenADAS parser

vsnever:
As discussed in #352 and #364, the user should be able to mix atomic data from different providers in a single local repository. In particular, some data such as the dictionary of spectral line wavelengths and Gaunt factors for bremsstrahlung, which are already distributed, or will be distributed with Cherab, are not taken from OpenADAS. Therefore, following @CnlPepper's [advise](https://github.com/cherab/core/issues/364#issuecomment-1065394097), this PR splits the old `opeanadas` module into two components: the `atomic` module for working with a local json repository, and the new `openadas` module for parsing ADAS files and installing them in a local repository.

The list of changes includes:

1. `openadas.repository` is moved to `atomic.repository`,
2. `openadas.rates` is moved to `atomic.rates`,
3. `openadas/openadas.py` is moved to `atomic/atomicdata.py` and `OpenADAS()` class is renamed to `AtomicData()`.
4. Default path to atomic data repository changed to `~/.cherab/atomicdata/default_repository`.
5. For backward compatibility, `openadas.repository`, `openadas.rates` and `openadas.OpenADAS()` now redirect to `atomic.repository`, `atomic.rates` and `atomic.AtomicData()` respectively.
6. All `AtomicData()` methods now search for the isotope data, and fallback to the element data only if the isotope data is missing. This behaviour is controlled by the `rate_element_fallback` argument, which is set to `True` by default.
7. Docstrings of the functions in the `atomic.repository` and `openadas.install` are improved. The new sections "Atomic data repository" and "Open-ADAS" are added to the documentation under the "Atomic Data" section.
8. All demos are updated to use the new `AtomicData()` interface.

Of course, the format for representing the atomic rates in the local repository is very specific to ADAS. In the future, this format can be generalised. For example, parameters such as interpolation type or the order of coordinates for interpolation should also be stored in json. I think that a fairly general format for representing atomic and molecular data in ITER IMAS, `AMNS_data` IDS, can serve as a guide here.

P.S. I used `git mv` for all moved files, but for some reason (too many changes?) git treats some of those files as new.


vsnever:
I've added support for the `CHERAB_ATOMIC_DATA` environment variable from #416.

The [PR #5](https://github.com/cherab/adas/pull/5) in `cherab-adas` demonstrates how much simpler the subpackage will become after this PR is accepted.

Since #400 and #415 add new atomic data, I will need to make additional changes after they are accepted.

vsnever:
Obviously, this change is too big to make in one PR. I'll break it down into several PRs. The first will affect only the documentation, improving docstrings of the `openadas.repository` and `openadas.install` functions and adding respective sections to the docs. The second will split the `openadas` module into the atomic data repository and the OpenADAS parser modules without adding new functionality. Subsequent PRs will add functions for data manipulation that are not included in OpenADAS and distributed with Cherab.

PR #379, opened 2022-09-13
Add Raysect 0.8 compatibility

vsnever:
Raysect 0.8 has been released, so since the development branch should build against Raysect's master, the `raysect-0.8-compat` branch is ready to be merged into development.


vsnever:
Yes, that's probably right. I'll leave this PR open for the future though.

jacklovell:
Now that Cherab 1.4.0 is out I think we can look at updating the development branch for raysect 0.8. Tests all pass: any objections to the merge?

Mateasek (approved):

Mateasek:
not from my side, thanks

vsnever:
Just one thing, can we temporarily specify that we need `matplotlib<=3.5.3`, because [this](https://github.com/raysect/source/issues/383) `set_window_title()` issue was fixed only in Raysect 0.8.1, which is not yet released?

jacklovell:
Not too keen on working around raysect's bugs, but in this case setting a limit on the matplotlib version does seem like a pragmatic solution. I'll make the change.

CnlPepper:
I'll release Raysect v0.8.1 this weekend, if I get time.

jacklovell:
Thanks @CnlPepper. In that case I propose we wait until 0.8.1 is out and then just support that going forwards. I can't see anything in the 0.8.1 changelog that would affect Cherab: there are no instances of `extract_translation` or `extract_rotation` in Cherab's source code.

CnlPepper:
@jacklovell I've released Raysect 0v.8.1. It fixes a number of niggling issues, such as the deprecated matplotlib window title call.

vsnever:
@jacklovell, should we add the `-O3` compilation flag like Raysect 0.8.1 does?

CnlPepper:
I'd recommend you do. Anaconda and other python distributions seem to have changed their defaults to -o2, which results in a performance drop.

jacklovell:
Interesting. Manylinux still uses `-O3` which is what I build the PyPI wheels on, so forcing the use of `-O3` in setup.py shouldn't make a difference for most users. I'll make this change and then merge.

Issue #292, opened 2021-04-30
Interface for external beam definition


mnurgaliev:
Dear Cherab team!

I would like to use Cherab framework for synthetic CXRS and BES diagnostics development. Initially, I planned to use the [FIDASIM](http://d3denergetic.github.io/FIDASIM/page/index.html) code to calculate the beam penetration into plasma but I did not find the solution for the setting up beam density externally. For the start I can use the `SingleRayAttenuator()`  but in perspective it would be great to have distribution of excited beam states, fast ion, and halo atoms, which are calculated in FIDASIM. What is the best way to make some sort of interface between FIDASIM output files (HDF5 format, [x,y,z] grid) and Cherab?

Thanks,
Maxim Nurgaliev

Mateasek:
Hi @mnurgaliev,

I would like to expand the beam models in near future and adding the possibility to add the beam distributions from external sources is on the list. I hope I can get to it in next few months. The fast ions and beam halo distributions are probably better to be defined through plasma species I would say. Cherab can support arbitrary 6D distributions, not only Maxwellian, so it can cover fast particles.

mnurgaliev:
Hi @Mateasek!

Thank you for the answer. Hope to see this feature soon. I agree about fast ions and beam halo, they can be defined through plasma species.

Mateasek:
I'm coming back to this issue, because it rose in other discussions. I think that to allow the external sources of beam properties, a rewrite of the beam architecture is needed, which might lead to backwards incompatibility.

The limiting factor now is I think the entanglement `Beam` class  with the `BeamAttenuator` class and the `_generate_geometry` method. I would like to change the architecture to make it closer to how the Thomson Scattering laser description now works.

The changes I propose are:

- Separation of the `Beam` class from any relation to spatial properties of the beam
- Creating a new class `BeamProfile`, which would take care of all the beam related properties as energy, density etc.
- `BeamProfile` would use `DistributionFunction` to describe the beam distribution in 6D
- `BeamProfile` would specify the beam primitives, based on its actual shape (which allows better encapsulation of the beam)

The impacts I see are

**advantages**:
- It will in my opinion offer a much simpler way to creating new beam models both general and ones tailored to specific usage, it will also allow to expand the possible levels of detail of beam description and emission models.
- It will be much simpler to get a new beam model in Cherab, for example MultiRayAttenuator.
- It will be possible to use external definition, because users would have to deal only with the proposed `BeamProfile` class. 
[...]

PR #383, opened 2022-10-26
Draft: New beam architecture

Mateasek:
This is the work in progress of the new beam architecture (mentioned in #292). There still can be major changes, please use the branch with caution. I wanted to publish it to get it out there so it can be used by pioneers and tested to collect suggestions.

### Changes Summary
**BeamDistribution**
Added BeamDistribution class, with DistributionFunction as baseclass. BeamDistribution handles spatial properties of beam, e.g. geometry, density distribution etc. (these were removed from Beam node). BeamDistribution will allow BeamModels to adress arbitrary level of complexity, from the most complicated 6D integration during observation. Simpler realisation of BeamModels (the currently available ones) can use mehods of the class providing the moments of the distribution: bulk_velocity, temperature, density.

**SingleRayAttenuator**
Was made more stand-alone. It now has _transform and _transform_inv attributes which are transforms from and to beam frame. This is to allow simple re-using of SingleRayAttenuator for also "MultiRayAttenuator" BeamDistribuions

**ThinBeam**
is a replacement of the current beam model.

**BeamModels** and **BeamMaterial** changed to work with the ThinBeam model and changes brought by the introduction of the BeamDistribution

### DIFF
--- a/cherab/core/model/beam/__init__.py
+++ b/cherab/core/model/beam/__init__.py
@@ -21,3 +21,4 @@
 
 from .charge_exchange import BeamCXLine
 from .beam_emission import BeamEmissionLine
+from .thin_beam import ThinBeam
--- a/cherab/core/model/beam/thin_beam.pyx
+++ b/cherab/core/model/beam/thin_beam.pyx
@@ -308,14 +308,12 @@ cdef class ThinBeam(BeamDistribution):
         # there must be an attenuator present to configure
         if not self._attenuator:
             return
-
-        # check necessary data is available
+        # there must be plasma present to configure
         if not self._plasma:
-            raise ValueError('The distribution must have a reference to a plasma object to be used with an attenuator.')
-
+            return
+        # there must be atomic_data present to configure
         if not self._atomic_data:
-            raise ValueError('The distribution must have an atomic data source to be used with an emission model.')
-
+            return
 
         # setup attenuator
         self._attenuator.distribution = self
--- a/demos/beam.py
+++ b/demos/beam.py
@@ -31,7 +31,7 @@ from raysect.optical.material.emitter import Checkerboard
 from cherab.core import Plasma, Beam, Species, Maxwellian
 from cherab.core.atomic import elements, Line
 from cherab.openadas import OpenADAS
-from cherab.core.model import SingleRayAttenuator, BeamCXLine
+from cherab.core.model import SingleRayAttenuator, BeamCXLine, ThinBeam
 from cherab.tools.plasmas import GaussianVolume
 
 integration_step = 0.02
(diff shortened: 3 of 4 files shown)


vsnever:
Hi @Mateasek,
Thank you very much for this draft PR. I agree that this is a major step in the right direction.

I looked at the output of the FEDASIM code, there neutral density is resolved over excited states. I believe other codes that calculate beam attenuation also output state-resolved neutral density. If only a state-unresolved distribution is supported in Cherab, we run into a data inconsistency problem when the beam attenuation is calculated by the external code with one collisional radiative model (CRM) and the radiation is calculated in Cherab with another CRM. 

I'm still thinking how the state-resolved distribution of neutrals can be implemented in the `Beam`. Perhaps, the beam should have a list of `(state, distribution)` pairs similar to `(species, distribution)` pairs in Plasma composition. Having the state-resolved distribution of neutrals should simplify the `BeamCXLine` and `BeamEmissionLine` model, but when the beam attenuation is calculated internally in Cherab, the calculation of beam population should be moved from the `BeamCXLine` to somewhere else.

Mateasek:
Hi @vsnever , thanks for the feedback. At the beginning I thought that the beam would have only a single particle species needed for the models to work. You are right, that more advanced models would probably need more beam species. In this case I think we should adopt something similar to the plasma composition and apply it to the beam. We should also implement state resolved model into the atomic model of cherab. Sorting this out only within the beam would be I think a bit dirty.

vsnever:
For the plasma composition, I think we could restrict ourselves to level resolution for hydrogen and H-like ions, and metastable resolution for other species. This would allow to use metastable-resolved PECs and Einstein coefficients for H and H-like ions, when the population density is calculated by an external code, e.g. EIRENE. That would be a huge step forward, but given the amount of change, do we have enough manpower?

vsnever:
I was thinking about how to add this new beam architecture in 1.5 without breaking user code and I came up with a possible solution.

I think, we should keep the beam properties, now associated with the `ThinBeam` distribution, until the next release. Beam distribution, if not explicitly set, should default to `ThinBeam`. So, these deprecated `Beam` properties should work as follows:
```python
    @property
    def divergence_x(self):
        if isinstance(self._distribution, ThinBeam):
            return self._distribution.divergence_x
        raise TypeError("Unable to get 'divergence_x', not a 'ThinBeam'.")

    @divergence_x.setter
    def divergence_x(self, double value):
        if isinstance(self._distribution, ThinBeam):
            warnings.warn("Setting beam distribution properties directly through a 'Beam' instance is deprecated and will be removed in the next release. Instead, use the 'BeamDistribution' instance properties.", DeprecationWarning)
            self._distribution.divergence_x = value
            self.notifier.notify()
        else:
            raise TypeError("Unable to set 'divergence_x', not a 'ThinBeam'.")

    @property
    def element(self):
        return self._distribution.element

    @element.setter
    def element(self, Element value not None):
        warnings.warn("Setting beam distribution properties directly through a 'Beam' instance is deprecated and will be removed in the next release. Instead, use the 'BeamDistribution' instance properties.", DeprecationWarning)
        self._distribution.element = value
```

The line `warnings.simplefilter('once', DeprecationWarning)` can be added at the top of the file to make these warnings appear only once.

Mateasek:
Hi @vsnever ,

at the end, we will add the new architecture along the old one. The new one will be sitting in `particle_beams`. This was suggested by @jacklovell during a meeting I had with him when I was visiting JET. I should update this in the issue at least, sorry. We will run the old and new architectures alongside for some time, add deprecation warnings to the old one and when we raise the major version, we will remove the old beam. This is I think the best compromise because we won't have to be introducing some non-optimal aspects into the new architecture.

I'm not in favour of the name "focal length" in this case . It is something the ion lense in the beam source has, but when you look at the generated beam profile the actual beam waist can be in some distance from the focal point of the lense. The difference depends largely on the source properties as beamlet divergences, focal length, etc. In case of a model which is based on the description of the beam itself I would like to use terms "beam waist" and "beam divergence" which is taken from the Gaussian model.

My plan is to add also a model of a beam based on individual bealmlets which will have a focal point into which all the beamlets aim.

PR #385, opened 2022-11-07
Fix Bremsstrahlung trapezium indexing

### LINKED ISSUE #384: Bremsstrahlung trapezium rule bug
I think that here we have a bug in the trapezium evaluation, don't we?

https://github.com/cherab/core/blob/c47a09ce7687bb01ebaca59e542924a28913b01a/cherab/core/model/plasma/bremsstrahlung.pyx#L77

I think that there should be (i + 1) term here, otherwise for the 1st iteration the `lower_wavelength` and `upper_wavelength` are equal, aren't they?

The line should look like:
` upper_wavelength = spectrum.min_wavelength + spectrum.delta_wavelength * (i + 1)`

Mateasek:
Fix indexing issue (#384)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ New:
 
 Bug Fixes:
 ----------
+* Fixed Bremsstrahlung trapezium evaluation (#384).
 * Fixed generomak plasma edge data paths.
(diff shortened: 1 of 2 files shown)


vsnever (approved):
Thanks for the fix, @Mateasek.

Issue #387, opened 2022-11-22
Add CylindricalMapper and PeriodicMapperXD


vsnever:
The `AxisymmetricMapper` maps the 2D function defined in the RZ-coordinates to Cartesian coordinates. However, Cherab does not currently have a function to map a 3D function defined in cylindrical coordinates to Cartesian coordinates. `CylindricalMapper` should fill this gap. The function being mapped is considered to be specified in the interval [0, 2π) in polar angle.

3D plasma simulation in tokamaks is often done for a toroidal sector, assuming periodicity in the toroidal direction. To map the results of such simulations to a full torus, a periodic mapper is needed. For generality, periodic mappers can be defined for 1D, 2D and 3D functions:
```python
PeriodicMapper1D(function1d, period)
PeriodicMapper2D(function2d, period_x, period_y)
PeriodicMapper3D(function3d, period_x, period_y, period_z)
```
Here it is assumed that functions being mapped are defined in [0, period) intervals. `period=None` means no periodicity in a given direction.

Then, to map a function defined in cylindrical coordinates on a 20° toroidal sector, the user can do:
```python
CylindricalMapper(PeriodicMapper3D(function, None, np.pi / 9, None))
```

This can be extended on vector functions as well.

vsnever:
Implemented in #388.

PR #388, opened 2022-11-30
Add cylindrical and periodic mappers

vsnever:
This solves #387 by extending the collection of Cherab mappers with `PeriodicMapper#D`, `VectorPeriodicMapper#D`, `CylindricalMapper` and `VectorCylindricalMapper`.

The periodic mappers are needed to handle the results of simulations carried out with periodic boundary conditions, while the cylindrical mappers are needed to handle the functions defined in cylindrical coordinates.

For example, with the new mappers a 3D function defined in cylindrical coordinates on a 20° toroidal sector can be mapped with:
```python
CylindricalMapper(PeriodicMapper3D(function, 0, np.pi / 9, 0))
```
and a vector function can be mapped with:
```python
VectorCylindricalMapper(VectorPeriodicMapper3D(vector_function, 0, np.pi / 9, 0))
```


vsnever:
I agree, let's move these to `cherab.math.transform` and rename to `CylindricalTransform`, `PeriodicTransformXD` etc.

vsnever:
The functions are renamed to `PeriodicTransform#D`, `VectorPeriodicTransform#D`, `CylindricalTransform` and `VectorCylindricalTransform` and moved to `cherab.math.transform`. The docs are updated to include the new `cherab.math.transform` sub-module.

jacklovell:
Looks good. Can you add unit tests for the `VectorPeriodic*Transform` classes? The scalar classes have tests but the vector classes are missing them.

(By the way, test coverage can be checked by installing recent versions of Cython and Coverage and running `./dev/build.sh --line-profile && coverage run -m unittest && coverage report` (with optional `coverage html` for easier navigation).

vsnever:
Thanks for the review, @jacklovell. I've added tests for the `VectorPeriodicTransform` functions.

jacklovell:
Thanks. Looks good to me now.

PR #392, opened 2022-12-22
Release 1.4.0

jacklovell:


vsnever:
@jacklovell, I just noticed that "Bug fixes" section is repeated twice in changelog.md. Also, remove "Fix wavelength indexing in Bremsstrahlung emission model. (#352)" and "Fixed Bremsstrahlung trapezium evaluation (#384)." are fixing the same bug. Keep only the last one.

jacklovell:
Good catch. I'll fix that. I'll also go through the commit history and check that all the changes are accurately reflected in the changelog.

jacklovell:
Cherab 1.4.0rc1 has been uploaded to PyPI: sdist and manylinux wheels for Python 3.7, 3.8, 3.9 and 3.10. 3.10 is built on manylinux2014 as earlier ones couldn't build the wheel due to dependency issues: the others are built on manylinux1 to match Raysect.

Please test these by running `pip install cherab==1.4.0rc1` on as many python installations as you can: I've tested all 4 versions on the Culham Freia machines and they seem to work OK.

vsnever:
I tested it on the ITER cluster and everything works fine.

The only thing I noticed, `CHANGELOG.md` contains two identical entries: "Fixed generomak plasma edge data paths" under "Bug fixes" section.

Mateasek:
Sorry for the delay. I tested the releaset with major python distributions we use at COMPASS and all seems to be fine, so from my side I think we can go ahead.

PR #402, opened 2023-01-25
Replace the half-sum with numerical integration when calculating the bremsstrahlung spectrum

### LINKED ISSUE #395: Add numerical integration of bremsstrahlung over spectral bins
Since integrators are now added to Cherab, when calculating the bremsstrahlung spectrum, the half-sum of the values at the lower and upper ends of the spectral bin can be replaced by integration.

https://github.com/cherab/core/blob/365c211acc1a3bfc45b273ab52d03f593b1fc4e2/cherab/core/model/plasma/bremsstrahlung.pyx#L110-L122

### LINKED ISSUE #409: Bremsstrahlung formula reference
Currently, the reference for the bremsstrahlung formula is the [Beurskens ITER LIDAR](https://aip.scitation.org/doi/full/10.1063/1.2969097) article, which uses an unexplained constant and is not in SI units.
Looking into references I was able to find an article by [Kadota](https://iopscience.iop.org/article/10.1088/0029-5515/20/2/010/meta), where the formula is also presented. However, without derivation or explanation and it is written in some [form of CGS unit system](https://en.wikipedia.org/wiki/Centimetre%E2%80%93gram%E2%80%93second_system_of_units#Various_extensions_of_the_CGS_system_to_electromagnetism) as the vacuum permittivity is not present. This is the oldest English written reference. It is pointing to [Kontinuierliche Spektren](https://link.springer.com/chapter/10.1007/978-3-642-45861-3_2) (in German), where I hit the limits of my language skills.
My colleague recommended me [Hutchinson's Principles of Plasma Diagnostics](https://iopscience.iop.org/article/10.1088/0741-3335/44/12/701/meta) where the formula is derived and the constants are in SI units. Some information about the gaunt factor is also present. The downside is that it is written with respect to frequency. But this can be solved by substitution factor.

$$
P_{\mathrm{ff}} \approx \int_{f\mathrm{min}}^{f_\mathrm{max}} j_\mathrm{ff} (f) \mathrm{d}f = 
			\int_{\lambda_\mathrm{max}}^{\lambda_\mathrm{min}} - \frac{c}{\lambda^2} j_\mathrm{ff}\left(\frac{c}{\lambda}\right) \mathrm{d}\lambda =
			\int_{\lambda_\mathrm{min}}^{\lambda_\mathrm{max}} \frac{c}{\lambda^2} j_\mathrm{ff}\left(\frac{c}{\lambda}\right) \mathrm{d}\lambda
$$

I did simple test and the bremsstrahlung based on Hutchinson was 1.008937... times greater than that computed using current formula. This seems to be caused by the limited accuracy of the constant in Beurskens article $0.95\cdot10^{−19}$.

As the difference is less than a percent, I would propose to at least include Hutchinson as a reference to the documentation for better clarity and optionally implement the equation with pre_factor based on fundamental constants or change it to be based on Hutchinson's book.

vsnever:
This fixes #395 by adding `integrator` argument to the `Bremsstrahlung` emission model. The new `BremsFunction(Function1D)` class is added for compatibility with `Integrator1D`.

**Update.** This also fixes #409 by replacing the numerical constant with an exact expression and changing the reference to bremsstrahlung equation in the docstring.


jacklovell:
Would be good to have test coverage of this too.

PR #404, opened 2023-02-09
Add analytic plasma demo using raysect's function framework

jacklovell:
This is roughly 25x faster than the pure python, class-based demo.

Note that to build the docs for this requires installing the `sphinx-tabs` package from PyPI. `sphinx` and `sphinx-rtd-theme` are already additional requirements for building the docs: I wonder if we should define this somewhere (either in optional dependencies in setup.py or in a requirements-docs.txt).

### DIFF
--- a/demos/plasmas/analytic_plasma_function_framework.py
+++ b/demos/plasmas/analytic_plasma_function_framework.py
@@ -1,0 +1,165 @@
+
+# Copyright 2016-2018 Euratom
+# Copyright 2016-2018 United Kingdom Atomic Energy Authority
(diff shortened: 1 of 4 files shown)


jacklovell:
Regarding the documentation dependencies, the [pypa installer](https://github.com/pypa/installer) project has quite an elegant solution to this by providing a `requirements.txt` file in the `doc` folder, specifically for requirements when building documentation.

jacklovell:
I decided in the end to go with an `extras_require` in `setup.py`: that way there doesn't need to be two separate pip invocations to install the package and the documentation dependencies.

PR #407, opened 2023-02-23
Build extensions in parallel by default

jacklovell:
This speeds up installation from source using pip, where it is otherwise cumbersome to pass the `--parallel` option to `setup.py build_ext`.


jacklovell:
@CnlPepper this would also apply to https://github.com/raysect/source/issues/422

PR #408, opened 2023-02-23
Package demo scripts and data in sdists and wheels

jacklovell:
This enables users who haven't cloned the git repository to still run the demos. Use distutils/setuptools' support for `data_files` to achieve this.

Also remove MANIFEST.in and `include_package_data=True`, use setuptools' `find_packages` and `package_data` instead. This allows finer control over what gets packaged, and fixes a deprecation warning in recent setuptools versions (see
https://github.com/pypa/setuptools/pull/3308 for details). It also prevents Cython-transpiled C source files being included in the wheel distribution, which roughly halves the size of the installed distribution.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,0 @@
-include CHANGELOG.md CITE.md AUTHORS.md
-include cherab/core/VERSION
-recursive-include cherab *.pyx *.pxd *.json *.cl *.npy *.obj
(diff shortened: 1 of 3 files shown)


vsnever (approved):
Thanks, @jacklovell. I've checked that it works on Mac as well. On Mac, the demos are installed in Library/Python/3.xx/share instead of .local/share.

PR #413, opened 2023-05-02
Add 'units' attribute to RayTransferPipeline#D

### LINKED ISSUE #412: Add 'units' attribute to RayTransferPipelines
Currently, `RayTransferPipeline#D` multiplies the calculated matrix by the sensitivity of the detector, which has units of [m<sup>2</sup> sr], but is equal to 1 for some non-physical observers (`PinholeCamera`, `VectorCamera`, `SightLine`). This can confuse a user who has switched from `SightLine` to `FibreOptic` or from `PinholeCamera` to `CCDArray` and expects the ray transfer matrix to still be calculated in [m] but receives it in [m<sup>3</sup> sr].

So, I suggest adding an attribute `units` to `RayTransferPipeline#D`, which can be either 'power' or 'radiance', similar to the Power and Radiance pipelines. This attribute determines whether the matrix is multiplied by sensitivity ('power') or not ('radiance'). The 'power' option will be used by default, so as not to break user code.

vsnever:
This PR fixes #412 by adding `units` attribute to `RayTransferPipeline#D`, which can be either 'power' or 'radiance', similar to the Power and Radiance pipelines. This attribute determines whether the ray transfer matrix is multiplied by the sensitivity of the detector ('power') or not ('radiance').

The `RayTransferPixelProcessor` class is now split into two subclasses: `PowerRayTransferPixelProcessor` and `RadianceRayTransferPixelProcessor`.

The `RayTransferPipeline#D` classes now inherit from both `Pipeline#D` classes and the new `RayTransferPipelineBase` class.

The tests for `RayTransferPipeline#D` are added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ New:
 * Support Raysect 0.8
 * Add PeriodicTransformXD and VectorPeriodicTransformXD functions to support the data simulated with periodic boundary conditions. (#387)
 * Add CylindricalTransform and VectorCylindricalTransform to transform functions from cylindrical to Cartesian coordinates. (#387)
-* Add the units attribute to RayTransferPipelineXD that determines whether the ray transfer matrix is multiplied by sensitivity or not. (#412)
+* Add the kind attribute to RayTransferPipelineXD that determines whether the ray transfer matrix is multiplied by sensitivity ('power') or not ('radiance'). (#412)
 
 
 Release 1.4.0 (3 Feb 2023)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,11 @@ Project Changelog
 Release 1.5.0 (TBD)
 -------------------
 
-Bug fixes:
-* Fix improper multiplication by sensitivity factor in RayTransferPixelProcessor resulting in wrong units (m^3 sr instead of m) of ray transfer matrices. (#412)
-
 New:
 * Support Raysect 0.8
 * Add PeriodicTransformXD and VectorPeriodicTransformXD functions to support the data simulated with periodic boundary conditions. (#387)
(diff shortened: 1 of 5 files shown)


skuba31:
Hi Vlad, thanks for handling this. It is one of the things I had to figure out when I was starting to use the ray transfer matrix over the single line integrated geometry/contribution matrices I was calculating with my own code. 

I suppose it is mostly personal preference but the name `units` feels to be slightly misleading as the units of ray transfer matrix are not 'radiance' in case of [m] or 'power' for [m<sup>3</sup> sr].

I personally would thinking about such property as something like `quantity` as the `power` or `radiance` is the result obtained after the matrix multiplication with discretised "power density" (W m<sup>-3</sup>sr<sup>-1</sup>). If the name `units` is to be used, I would set the accepted values to be 'length' and either of 'throughput' or 'volume-angle' or 'volume-sr'...

vsnever:
Hi Jakub, the attribute is badly named indeed, thanks. However, I'd like to keep these `power`/`radiance` options because they match the Raysect pipelines and are easy to remember. I suggest renaming the `units` attribute to `kind`. The `quantity` is better than `units`, but still the power density or the radiance are the quantities of the matrix product, not the matrix itself.

skuba31:
I agree, `kind` is better name for the attribute.

jacklovell (approved):
Looks good to me, thanks. 

One thing to be careful of is that "radiance" is the average radiance over the entire solid angle seen by the observer. For aperture/detector systems like bolometers (and other relatives of TargettedPixel) the radiance is therefore lower than what you would typically calculate when taking only the solid angle subtended by the aperture. This is documented in the bolometry section of the documentation, but other observers which would be more likely to use the RTM framework may benefit from having this pointed out if it applies to them too. Let's wait and see whether this becomes an issue for end users or not.

PR #415, opened 2023-05-14
Update BeamEmissionMultiplet (MSE) model

### LINKED ISSUE #410: Update BeamEmissionMultiplet with more accurate MSE calculation
The spectrum calculation in the current BeamEmissionMultiplet class uses constant line intensity ratios, but in a real situation the ratios also depend on the observation direction. Better intensity ratios, or a method to calculate them might also be available.

vsnever:
This is a draft PR with an update for the `BeamEmissionMultiplet` line shape, which models the MSE under pure linear Stark conditions. This PR is created on top of the PR #400 and contains the following changes:

- The `BeamEmissionLine` model now supports custom line shapes to make it easier to add other MSE line shape models in the future (e.g. the **E^B** model).
- The σ1/σ0, π2/ π3 and π4/ π3 ratios in `BeamEmissionMultiplet` now the functions of beam energy and electron density and are thus compatible with the data from O. Marchuk's [paper](http://dx.doi.org/10.1088/0953-4075/43/1/011002).
- The `BeamEmissionMultiplet` now returns the spectrum with respect to the angle between the ray and **E**=**v^B**.


vsnever:
My colleagues recently performed calculations similar to those in O. Marchuk's [paper](http://dx.doi.org/10.1088/0953-4075/43/1/011002) and obtained the dependences of the Balmer-alpha MSE spectrum component ratios on the beam energy, electron density and magnetic field strength. The paper by A.V. Demura, D.S. Leontiev and V.S. Lisitsa entitled "Polarization characteristics of electrodynamic Stark effect" is accepted to J. Exp. Theor. Phys and will be published early next year. They also agreed to provide the tabulated data for the Balmer-alpha MSE components (σ0, σ1, π2, π3 and π4) for Cherab and Dmitry Leontiev performed calculations for a broader range of parameters then in the paper.

There is a discrepancy between the results by _O. Marchuk_ and _A.V. Demura et al_ shown on the following plots for the beam energies of 100 keV and 40 keV.

Demura_Marchuk_compare_100keV
Demura_Marchuk_compare_40keV


O. Marchuk's results are likely to be more accurate, because Glauber approximation used in his paper should provide more accurate results compared to Born approximation used in _A.V. Demura et al_. However, currently in Cherab we only have constant values for the MSE component ratios, so considering that we have no permission to include O. Marchuk's data, I think replacing these constant values with tabulated data for different beam energies, electron densities and magnetic field strengths from _A.V. Demura et al_ would be a huge step forward.

Here is a comparison of the old MSE spectrum model (constant ratios, no dependency on the angle between the ray and **E=v^B**) with the new one when using data from _A.V. Demura et al_ or _O. Marchuk_.

Demura_Marchuk_spectra_compare

The beam and LoS properties for this plot are the same as in the `beam_emission_spectrum.py` demo, but the `peak_density` of the plasma slab was set to 10<sup>19</sup> m<sup>-3</sup> and `beam_temperature` to 5 eV.

In #377, default atomic data will be moved to a local repository, so that the user can replace it with O. Marchuk's data if he/she has access to, for example, M. von Hellermann's "Simulation of Spectra" code.

PR #418, opened 2023-08-09
Fix NotImplementedError in the TestLaserSpectrum unit test.

### LINKED ISSUE #417: NotImplementedError in TestLaserSpectrum unit test
The `TestLaserSpectrum` unit test raises `NotImplementedError` with Cython 3.0 because the `evaluate()` method is not implemented in the base `LaserSpectrum` class. For some reason this exception was ignored with Cython 0.29.

I suggest doing the tests for the `ConstantSpectrum` as it is the simplest `LaserSpectrum` subclass that implements the `evaluate()` method.

vsnever:
This fixes #417 by replacing the base `TestLaserSpectrum` with its simplest subclass `ConstantSpectrum`. It also fixes the test for the `bins` property.

### DIFF
--- a/cherab/core/laser/tests/test_laserspectrum.py
+++ b/cherab/core/laser/tests/test_laserspectrum.py
@@ -1,8 +1,10 @@
 import unittest
 import numpy as np
 
-from cherab.core.laser.laserspectrum import LaserSpectrum
+# The evaluate() method is not implemented in the base LaserSpectrum, so importing the simplest subclass.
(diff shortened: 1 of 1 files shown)


Mateasek (approved):
Thanks for fixing it @vsnever .  I didn't run the tests myself, but the code seems ok to me.

vsnever:
CI tests failed on Python 3.10 because Raysect 0.8.1 does not have a wheel for 3.10 and is not compatible with Cython 3.0.

vsnever:
Thanks @Mateasek, I'll merge it then.

Issue #420, opened 2023-10-26
Laser material transformation cache not updated properly


Mateasek:
The [caches of affine transforms](https://github.com/cherab/core/blob/547e7d3b5ae979ae4bf7403c6ba1893c66b80bd3/cherab/core/laser/material.pyx#L35C2-L36C63) in the material of laser segments are not [updated properly](https://github.com/cherab/core/blob/547e7d3b5ae979ae4bf7403c6ba1893c66b80bd3/cherab/core/laser/node.pyx#L243) after [change of laser transform](https://github.com/cherab/core/blob/547e7d3b5ae979ae4bf7403c6ba1893c66b80bd3/cherab/core/laser/node.pyx#L260).

I will try to describe what I think caused the problem, to have a note in case I end up doing the same mistake. I don't expect anyone to verify.

**Intended behaviour**

- Laser geometry is specified by LaserProfie and the currently available profiles use segmented beams (i.e. instead of using one long cylinder, the laser cylindrical segments)
- The Laser node stores the laser primitives as its children in the scenegraph
- The laser material caches affine transforms to plasma and laser nodes to speed up observations
- The material of each laser primitive has to be updated after the Laser node transform changes which is done by assigning a new LaserMaterial to every segment
- The reassigning of the LaserMaterials is initiated by the [_modified](https://github.com/cherab/core/blob/547e7d3b5ae979ae4bf7403c6ba1893c66b80bd3/cherab/core/laser/node.pyx#L260) method of Laser node being called as a reaction to transform change by Raysect's notification system
- the _modified method calls the [_configure_material](https://github.com/cherab/core/blob/547e7d3b5ae979ae4bf7403c6ba1893c66b80bd3/cherab/core/laser/node.pyx#L243) method which assigns new materials to every laser primitive.

**Possible source or problem**
- The caching of transforms in the material uses transform of the laser primitives
- laser primitives are one level lower than the Laser node in the scenegraph (are children of Laser)
- when Laser.transform is change, the notification system is triggered and the notification reaches the Laser._modified method earlier than the notification to the laser primitives to change their to_root transforms
- The result is the transform caches in the new [LaserMaterial](https://github.com/cherab/core/blob/547e7d3b5ae979ae4bf7403c6ba1893c66b80bd3/cherab/core/laser/material.pyx#L29) instances use the deprecated transforms, because the laser primitives were not yet updated.

[...]

PR #421, opened 2023-10-26
Fix update bug of LaserMaterial

Mateasek:
Fixes problem described in #420 

In short:
Laser primitives are now dropped and recreated after Laser.transform changes which removes the problem of old transfomrs being cached in LaserMaterial

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ New:
 * Add the kind attribute to RayTransferPipelineXD that determines whether the ray transfer matrix is multiplied by sensitivity ('power') or not ('radiance'). (#412)
 
 Bug fixes:
-* Fix deprecated cached transforms in LaserMaterial after laser.transform update
+* Fix deprecated transforms being cached in LaserMaterial after laser.transform update (#420)
 
 Release 1.4.0 (3 Feb 2023)
 -------------------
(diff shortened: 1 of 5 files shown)


vsnever (approved):
Hi @Mateasek, nice catch!

Ideally, if a material caches geometry of the primitive, any change in the geometry should automatically trigger cache rebuild. However, I don't know how to implement this, because laser segments are just basic primitives with an empty `_modified()` method. Considering this, your solution is fine.

vsnever:
Just as an option, wouldn't it be safer in general to cache affine matrices the first time the `emission_function()` is called, rather than when the `LaserMaterial` is initialised? In this case, the  `LaserMaterial` must keep the references to laser and laser segment nodes, but the `Laser` does not have to rebuild its internal geometry when the node is transformed.

Mateasek:
Hi @vsnever 

thanks for looking into it. The reaction to the material change would mean that the laser primitives couldn't be regular Raysect primitives, which I would like to avoid.

I like the idea of caching the transforms in the emission function. It is indeed the safest option. Thanks for suggesting it. I will apply the changes. I don't think that rebuilding the primitives on every change is a big deal, but from the point of view of safety your idea is much better.

Mateasek:
As suggested by @vsnever the transforms are now checked and cached in the emission function.

Also I added a check for the number of primitives the material is assigned to. Because of the transform caces the material should be assigned to only a single primitive for safety.

vsnever (approved):
Thanks @Mateasek, I like this fix better. Could you please merge the development branch into this, because it now has a fix for #422?

Mateasek:
@vsnever, I merged the latest changes from development.

jacklovell (approved):
Looks OK to me. Thanks Matej!

PR #428, opened 2024-03-27
Fix IRVB calculate sensitivity method

jacklovell:
Spectral pipelines have no display_progress option.


jacklovell:
I thought about adding some test coverage for this, but it's going to be quite involved creating a minimal working example for unit tests that doesn't take ages to run while having good enough statistics for a meaningful test for correctness. Maybe when we have more Generomak observers we can think about adding some test coverage for the `cherab.tools.observers` classes.

vsnever (approved):
Looks fine to me. Thanks!

PR #430, opened 2024-04-04
Add support for Apple GPUs to get_flops() function.

### LINKED ISSUE #429: Unable to detect Apple GPU when using SartOpencl
The `get_flops()` function used in `get_best_gpu()` does not support the Apple GPU used in the Apple M SoC. Therefore, unless a device is manually passed into `SartOpencl`, the `select_device()` function will not be able to recognise the Apple GPU as a supported device.

This can be fixed by adding Apple to the list of supported device vendors.

vsnever:
This fixes #429 by providing performance estimates for Apple GPUs in the `get_flops()` function.

According to [wikipedia](https://en.wikipedia.org/wiki/Apple_silicon#Comparison_of_M_series_processors), Apple GPUs on the M1, M2 and M3 SoCs have 128 FP32 ALU lanes per core (compute unit).


jacklovell:
Looks OK to me. Thanks.

PR #432, opened 2024-04-27
Fix incorrect docstrings for atomic rate base classes

### LINKED ISSUE #431: Obsolete docstrings in the interface classes for atomic rates
The docstrings for the interface atomic rate classes are outdated and instruct the user with the wrong order of parameters for the `evaluate()` function. For example:

https://github.com/cherab/core/blob/bd36ca157f6524c8c10cb56a78f1a820995a5e85/cherab/core/atomic/rates.pyx#L23-L41

The units of the return values are also incorrect.

vsnever:
This updates the docstrings for the atomic process rate base classes and fixes #431.

### DIFF
--- a/cherab/core/atomic/rates.pxd
+++ b/cherab/core/atomic/rates.pxd
@@ -75,7 +75,7 @@ cdef class TotalRadiatedPower:
     cdef:
         readonly Element element
 
-    cdef double evaluate(self, double density, double temperature) except? -1e999
+    cdef double evaluate(self, double electron_density, double electron_temperature) except? -1e999
 
 
 cdef class _RadiatedPower:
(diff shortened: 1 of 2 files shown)


Mateasek (approved):
Seems good to me

jacklovell (changes requested):
Changing the names of the arguments is technically a breaking change for anybody passing arguments as keywords to these functions. I get that aesthetically it's nicer to have a uniform convention of `density` and `temperature` as names, but have you checked there's nowhere else in the code that calls these by keyword (may not be covered by existing unit tests)?

I'd be inclined to leave the argument names as they are and only update the docstrings.

vsnever:
Thanks for the reviews, @Mateasek, @jacklovell.

>Changing the names of the arguments is technically a breaking change for anybody passing arguments as keywords to these functions. I get that aesthetically it's nicer to have a uniform convention of density and temperature as names, but have you checked there's nowhere else in the code that calls these by keyword (may not be covered by existing unit tests)?
 I'd be inclined to leave the argument names as they are and only update the docstrings.
 
Yes, indeed, in Python positional arguments can also be passed by keyword, and changing the name of a positional argument can cause a `TypeError`. I did not think about that. I returned the arguments to their old names.

jacklovell (approved):

PR #433, opened 2024-05-02
Improve density and direction calculation in the neutral beam model

### LINKED ISSUE #414: The vector returned by Beam.direction(x, y, z) does not depend on coordinates
I was working on updating the MSE model and found that the calculation of the direction of the beam at a point, which is used as a direction of beam bulk velocity in the `BeamEmissionLine` and `BeamCXLine` models, is rather strange:
https://github.com/cherab/core/blob/547e7d3b5ae979ae4bf7403c6ba1893c66b80bd3/cherab/core/beam/node.pyx#L233-L253
The vector returned by this function is independent of the given x, y coordinates and always points in the same direction, determined by the beam's divergence angles.

I think that if the model takes into account beam divergence, then at least the polar angle should be taken into account:
```python
        # calculate direction from divergence
        cdef double phi = atan2(y, x)
        cdef double dx = cos(phi) * tan(DEGREES_TO_RADIANS * self._divergence_x)
        cdef double dy = sin(phi) * tan(DEGREES_TO_RADIANS * self._divergence_y)
        return new_vector3d(dx, dy, 1.0).normalise()
```
However, I'm not an NBI expert, so I'm asking for advice here. @Mateasek, what do you think?

vsnever:
This updates density and direction calculation of the neutral beam model and fixes #414.

1. The beam deviation is now calculated as $$\sigma_x(z) = \sqrt{\sigma^2 + (ztg(\alpha_x))^2}\hspace{0.5cm}\sigma_y(z) = \sqrt{\sigma^2 + (ztg(\alpha_y))^2},$$ which is consistent with the Gaussian beam model, where the beam deviation σ(z) due to both beam divergence and deviation at beam origin is the result of the convolution of the two Gaussian functions as explained [here](https://github.com/cherab/core/issues/414#issuecomment-1549412429). Previously, the beam deviation was modelled as: $$\sigma_x(z) =\sigma + ztg(\alpha_x)\hspace{0.5cm}\sigma_y(z) = \sigma + ztg(\alpha_y),$$ which is unclear to me.
2. In this case, the (non-normalised) direction of velocity of the beam atoms can be calculated as follows: $$e_x = x\frac{(ztg(\alpha_x))^2}{\sigma^2 + (ztg(\alpha_x))^2},$$ $$e_y = y\frac{(ztg(\alpha_y))^2}{\sigma^2 + (ztg(\alpha_y))^2},$$ $$e_z = z.$$ This gives the following pattern for the beam direction (σ=0.1 m and α=5°):
divergence

The broadening of the BES line shape is now naturally determined by the beam divergence. Therefore, setting beam temperature to the physically unreasonable values is no longer required. The parameters of the beam in the BES demo are updated.

Although this change does impact user results, I think it is justified by the improved consistency of the beam model.


Mateasek (changes requested):
Hi @vsnever,

thanks for the changes. The code itself looks clean and I couldn't find any problems. I would only suggest one improvement regarding the documentation. Since the beam direction became more complex, I think it would be nice to add the equations you use into the documentation. I think it would be enough if you add the equations and plot you posted in the opening post of this PR. It would be explanatory enough and would remove the necessity to go through the cython code for the users, what do you think?

jacklovell (changes requested):
Thanks @vsnever. I'm with @Mateasek: the changes are sound but it would be even better to explicitly document what calculation is being used.

I'd also suggest adding some emphasis (`**` or `*`) to the line in the changelog highlighting that this will change existing results.

If there aren't already any unit test associated with this, now is also a good time to add them. But don't worry if that's too big an project: I wouldn't want to hold up an otherwise useful change because of the work involved in setting up a mock beam for tests.

vsnever:
Thanks for the reviews, @Mateasek, @jacklovell.

> Since the beam direction became more complex, I think it would be nice to add the equations you use into the documentation. I think it would be enough if you add the equations and plot you posted in the opening post of this PR. It would be explanatory enough and would remove the necessity to go through the cython code for the users, what do you think?

I added equations for the beam density and direction in `SIngleRayAttenuator.density()` and `Beam.direction()` docstrings respectively. I noticed that in the docs, the description of the `SingleRayAttenuator` class was placed under the "CXS models" section, so I moved it to a separate "Beam attenuation" section. The beam direction plot is added to the `particle_beams.rst`.

> I'd also suggest adding some emphasis (** or *) to the line in the changelog highlighting that this will change existing results.

I agree, the line announcing this change in the changelog is now in bold.

> If there aren't already any unit test associated with this, now is also a good time to add them. But don't worry if that's too big a project: I wouldn't want to hold up an otherwise useful change because of the work involved in setting up a mock beam for tests.

We already have tests for the beam CX model, so setting up a mock beam and plasma is not a problem. I added the tests for the beam density and direction. They test these values on-axis, off-axis and outside the beam.

cherab/core/beam/node.pyx
```diff
@@ -232,7 +235,24 @@ cdef class Beam(Node):
 
     cpdef Vector3D direction(self, double x, double y, double z):
         """
```
jacklovell, line 237: Use `r"""` and get rid of double backslashes in math formulae.

cherab/core/model/attenuator/singleray.pyx
```diff
@@ -87,17 +102,40 @@ cdef class SingleRayAttenuator(BeamAttenuator):
     @cython.cdivision(True)
     cpdef double density(self, double x, double y, double z) except? -1e999:
         """
```
jacklovell, line 104: `r"""` again.

cherab/core/tests/test_beam.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, donor_metastable, value):
+        self.donor_metastable = donor_metastable
+        self.value = value
+
+    def evaluate(self, energy, density, temperature):
+
+        return self.value
+
+
+class TestAtomicData(AtomicData):
```
jacklovell, line 48: Rename this `MockAtomicData`: classes starting with `Test` are assumed by both unittest and pytest to contain test cases and this convention is broken here which in turn throws up warnings in the test discovery process.

jacklovell (changes requested):
Looks good. Couple of minor suggestions.

cherab/core/beam/node.pyx
vsnever: Done.

cherab/core/model/attenuator/singleray.pyx
vsnever: Done,

vsnever:
>Looks good. Couple of minor suggestions.

Thanks, @jacklovell. All done.

Mateasek (approved):
Thanks @vsnever for improving the docstrings, I have no further suggestions.

jacklovell (approved):

PR #434, opened 2024-05-09
Fix Sphinx warnings about documents not being included in toctree

jacklovell:
The demos are in fact not included in any toctree, but they are not intended to be so mark them as orphans as recommended in the Sphinx documentation.

These warnings make it harder to spot other - genuine - warnings, so removing them is useful for ensuring documentation is produced correctly.

### DIFF
--- a/docs/source/demonstrations/active_spectroscopy/beam_bes.rst
+++ b/docs/source/demonstrations/active_spectroscopy/beam_bes.rst
@@ -1,3 +1,5 @@
+:orphan:
+
 
 .. _beam_bes:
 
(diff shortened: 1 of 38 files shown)


vsnever (approved):
Thanks, @jacklovell. Indeed, it was difficult to find errors in the docs due to the large number of warnings. This fixed the problem for me.

There are three legacy files in the docs, presumably originating from a pre-release version of Cherab, also not included in the documentation tree, namely: `models/beam/beam_attenuation_calculation.rst`, `models/cxs/atomic_data_calculation.rst`, `plasmas/plasma_sources.rst`. However, I think that these three files should be left as is, even if they generate warnings, because otherwise we will forget about them, and they contain useful information that can be transferred to some other files in the future.

jacklovell:
Yes, I deliberately left those files in there. The information in them should be checked for accuracy and transferred into public documentation, but that's out of the scope of this particular PR.

PR #439, opened 2024-05-14
Rescale matrices and vectors in invert_regularised_nnls and allow kwargs.

### LINKED ISSUE #438: invert_regularised_nnls returns zero emissivity profiles in bolometer demos
It appears that the `nnls()` method is reimplemented in SciPy 1.12, and due to some changes in the termination criteria, `inversion_with_raytransfer.py`, `inversion_with_voxels.py` bolometer demos and `admit_tomographic_inversion.py` from #427 return zero emissivity profiles.

Normalising the `c_matrix` and `d_vector` by `d_vector.max()` before passing them to `nnls()`, solves the problem.

Also, the new `nnls` implementation is much slower if `atol` is not set. I suggest adding `**kwargs` to `invert_regularised_nnls()` so that the user can control `maxiter` and `atol` parameters.

vsnever:
This fixes #438 by rescaling the `w_matrix`, `b_vector` and `tikhonov_matrix` before passing them to `nnls` to avoid possible issues with the recent versions of SciPy. Also, `**kwargs` are added, so the user can control the `nnls` termination criteria.


Mateasek (approved):
This seems good code wise. Since I'm not familiar with tomographic inversions I can't judge if there are any other issues.

jacklovell:
Thanks for this @vsnever: we're still stuck on Python 3.7 here at Culham and so don't get to experience these Scipy "improvements" often. Accuracy and performance regressions in a library as well maintained as Scipy are concerning: I'd have expected better backwards compatibility unless we're somehow misusing the Scipy functions in Cherab[*].

Have you checked that this normalisation scheme gives identical results (at least to within numerical precision) as the original function, for Scipy < 1.12.0? Intuitively I feel like it should, but it'd be nice to get confirmation. I don't *think* we have any unit tests to check for regression in the matrix inversion tools.

[*] I'm beginning to think scipy.optimize.nnls is not the best routine to use for bolometry actually: since both the geometry matrix and the regularisation operator are sparse we could likely benefit from using some of Scipy's sparse-aware routines. I'll make a separate issue and take a look into this.

vsnever:
> Have you checked that this normalisation scheme gives identical results (at least to within numerical precision) as the original function, for Scipy < 1.12.0? Intuitively I feel like it should, but it'd be nice to get confirmation. I don't think we have any unit tests to check for regression in the matrix inversion tools.

Yes, I tested this. For older versions of SciPy, the results with and without normalisation are identical. However, I forgot to change the `rnorm` scale back. Now fixed.

jacklovell (approved):

PR #447, opened 2024-07-14
Require Cython 3 as a dependency

### LINKED ISSUE #446: Move to Cython 3
Cherab builds perfectly with Cython 3, but the build dependencies currently specify `cytnon >= 0.28`. Cherab additional modules require Cython 3 to compile. In addition, Cython 3 allows to get rid of annoying NumPy-1.7 C-API warnings.

I suggest specifying `cython ~= 3` in the build dependencies and runtime requirements.

vsnever:
This PR fixes #446.

- Cython 3 is listed as a build and runtime dependency. 
- CI workflows have also been updated.
- The C macros are added to get rid of the deprecated Numpy API warnings as specified in the [user guide](https://cython.readthedocs.io/en/3.0.x/src/userguide/source_files_and_compilation.html#configuring-the-c-build).
- `-Wno-unreachable-code-fallthrough` compilation flag is added to silence [false unreachable code warnings](https://github.com/cython/cython/issues/5041).


setup.py
```diff
@@ -115,14 +117,15 @@
     long_description=long_description,
     long_description_content_type="text/markdown",
     install_requires=[
+        "cython~=3.0",
```
jacklovell, line 120: Cython is not a runtime dependency, so doesn't need to be in `install_requires`. It's only needed for building.

jacklovell (changes requested):

Mateasek (commented):
I'm out of my depth here. I would ask @jacklovell and @vsnever to agree on this one.

setup.py
vsnever: Should I then remove it from `requirements.txt` too? Or is it useful to have it there when someone wants to install both build and runtime dependencies using `pip install -r requirements.txt`?

jacklovell: `requirements.txt` is for developers to set up a useful working environment, so the Cython dependency should remain in there. The `install_requires` section in `setup.py` should only specify the packages required to use an already-built wheel after it has been installed (it's not a great keyword name IMO).

vsnever: Thanks for the explanation. It's done.

jacklovell (approved):

PR #449, opened 2024-07-20
Tests: Move Plasma and Beam set up into the setUp() method

### LINKED ISSUE #448: Add setUp() to the test cases containing Plasma and/or Beam object
As suggested in https://github.com/cherab/core/pull/386#discussion_r1602799014, `Plasma`, `Beam`, or any other object whose state can be changed in the unit tests should be created and configured in the `setUp()` method to ensure their state is consistent between tests.

vsnever:
This fixes #448 by moving the creation and configuration of the `Plasma` and `Beam` to the `setUp()` method in the `TestBeam`, `TestBeamCXLine`, `TestBremsstrahlung` and `TestLineShapes` test cases.

### DIFF
--- a/cherab/core/tests/test_beam.py
+++ b/cherab/core/tests/test_beam.py
@@ -55,32 +55,38 @@ class MockAtomicData(AtomicData):
 
 class TestBeam(unittest.TestCase):
 
-    atomic_data = MockAtomicData()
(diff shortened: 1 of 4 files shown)


Mateasek (approved):
thanks for the improvement @vsnever

PR #456, opened 2024-08-02
Fix wrong naming in TestTotalRadiatedPower

vsnever:
This PR simply renames copy-pasted `test_beam_density()` to `test_total_radiated_power()` in the `TestTotalRadiatedPower`.


Mateasek (approved):

Mateasek:
I think there is no need to wait for @jacklovell with merging such a simple PR, right?

PR #461, opened 2024-08-04
Add r prefix to strings used in regular expression matching

### LINKED ISSUE #460: Use raw strings in regular expression matching when parsing ADAS files
Python 3.12 generates a lot of `SyntaxWarnings` related to ADF11 and ADF15 parsers when running any Cherab script. For example:
```
...cherab/openadas/parse/adf11.py:109: SyntaxWarning: invalid escape sequence '\s'
  iptr_pos = re.search("IPRT\s*=*\s*[0-9]+\s*", lines[i]).group()  # get the IPRT part
```

Therefore, raw strings should be used in regular expression matching.

vsnever:
This PR fixes #460 by adding _r_ prefix to strings used in regular expression matching when parsing ADF11 and ADF15 files.

### DIFF
--- a/cherab/openadas/parse/adf11.py
+++ b/cherab/openadas/parse/adf11.py
@@ -38,7 +38,7 @@ def parse_adf11(element, adf_file_path):
     with open(adf_file_path, "r") as source_file:
 
         lines = source_file.readlines()  # read file contents by lines
-        tmp = re.split("\s{2,}", lines[0].strip())  # split into relevant variables
+        tmp = re.split(r"\s{2,}", lines[0].strip())  # split into relevant variables
         # exctract variables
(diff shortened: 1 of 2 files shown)


jacklovell:
Thanks: looks good to me. We don't have test coverage for these ADAS parsers: can you confirm that the parsed data is identical to that obtained with 1.5.0rc1?

vsnever:
> Thanks: looks good to me. We don't have test coverage for these ADAS parsers: can you confirm that the parsed data is identical to that obtained with 1.5.0rc1?

The repositories created with `populate()` before and after adding _r_ prefix are identical.

jacklovell (approved):

Issue #465, opened 2024-08-30
Add electric field vector function to Plasma object


vsnever:
It is known that in some cases the radial electric field can affect MSE measurements. Ideally, the MSE model in Cherab should take into account static electric field, **E**, in addition to the Lorentz **v**⨉**B** term.

I propose adding the `e_field` property to the Plasma object by analogy with the `b_field`.

Mateasek:
Closed with #466

PR #466, opened 2024-08-30
Add e_field attribute to Plasma object for electric field vector

vsnever:
This PR implements #465 by introducing the `e_field` attribute to the `Plasma` object for the electric field vector function. This electric field will be used in the new `BeamEmissionMultiplet` model (#415) as an addition to the Lorentz field **v**⨉**B**.


jacklovell (approved):
Looks good to me. It would be nice to have a couple of simple unit tests to check setting and getting works (mainly useful for regression testing later on I expect).

Mateasek (approved):
Hi @vsnever 

I think this one is good to go. As @jacklovell said, you could add tests, but looking at the code I don't see many options here. You could check that the property value is set to whatever `autowrap_vectorfunction3d` is supposed to return when given a valid input or `None`.

Issue #471, opened 2025-01-13
Add Synchrotron Radiation model


Mateasek:
I would like to add a Synchrotron Radiation (SR) emission model to Cherab. The radiation will be based on the local properties of the electron distribution function. Because of the directivity of the radiation, I would like to provide users with a number of integrators suiting well wider range of speed and accuracy requirements. For this reason I will implement 2D integrators (raised in a separate issue).

TODO:

- [ ] radiated spectrum
- [ ] radiated power by a single electron
- [ ] directivity of the radiation
- [ ] 2D integration over polar coordinates in radiator's frame
- [ ] tests
- [ ] documentation

PR #473, opened 2025-01-13
Add Integrator2D base class

### LINKED ISSUE #472: Add 2D integrator
Add `Integrator2D` class to support #471 

To separate the commits, I would first add the base class for 2D integrators and then implementations. The base class will copy `Integrator1D` from @vsnever.

Mateasek:
This is the base class for `Integrator2D` proposed in #472 

I mostly copy-pasted the `Integrator1D` by @vsnever


vsnever:
Hi @Mateasek,

Thanks for this PR. The `Integrator2D` base class takes `y_lower` and `y_upper` as doubles and not as `Function1D` objects. In general, the boundaries for y should be a function of x. Does your interface cover all the integration cases in #471?

Mateasek:
Hi @vsnever,

>  In general, the boundaries for y should be a function of x. 

This is a good point, I will change it.

 > Does your interface cover all the integration cases in https://github.com/cherab/core/issues/471?

I hope so. The model I'm working on (at least for now) needs to integrate only over 2 dimensions. I simplify the radiated spectrum, otherwise it would need 3D integrator which for now seems unnecessary and too computationally demanding. If 2D is not enough, I can always add 3D integrators and we will have 2D for future usage in the worst case.

vsnever (approved):
I think it's perfect now! Thanks.

Mateasek:
Thanks for review @vsnever

Issue #475, opened 2025-01-17
Add 2D Gauss-Legendre quadrature integrator


Mateasek:
I would like to add a `GaussianQuadrature2D` integrator.  The implementation will be close to what @vsnever did in `GaussianQuadrature` class. It is related to issues #472 and #471 . I will separate the PR from #473 for clarity and to keep PR small.

jacklovell:
Fixed in #476

PR #476, opened 2025-01-17
Add GaussianQuadrature2D

Mateasek:
Adds `GaussianQuadrature2D` class mentioned in #475.

Adds also integrators (1D, 2D) to our documentation.

This PR will be marked as draft until PR #473 is merged.


cherab/core/math/integrators/__init__.pxd
```diff
@@ -17,5 +17,5 @@
 # under the Licence.
 
 from cherab.core.math.integrators.integrators1d cimport Integrator1D, GaussianQuadrature
```
jacklovell: For consistency, we should rename `GaussianQuadrature` to `GaussianQuadrature1D` to distinguish it from the new 2D class. But I suggest keeping `GaussianQuadrature` as an alias to maintain backwards compatibility for now, and deprecate it in another release or so (removed in Cherab 2.0).

cherab/core/math/integrators/__init__.py
```diff
@@ -17,4 +17,4 @@
 # under the Licence.
 
 from .integrators1d import Integrator1D, GaussianQuadrature
```
jacklovell: Ditto `GaussianQuadrature1D`

cherab/core/math/integrators/integrators2d.pxd
```diff
@@ -16,6 +16,8 @@
 # See the Licence for the specific language governing permissions and limitations
 # under the Licence.
 
+from numpy cimport ndarray
```
jacklovell: This is unnecessary: you don't make use of the typed `_x_roots`, `_x_weights`, `_y_roots` and `_y_weights` in any tight loops: they're only assigned to per the output of `roots_legendre`, which goes through the Python layer anyway. So you can cdef these as `object` instead of `ndarray` and save a numpy `cimport`, which will speed up compilation.

All the typed access is (correctly) done through the memoryviews ayway.

cherab/core/math/integrators/integrators2d.pyx
```diff
@@ -59,3 +65,305 @@ cdef class Integrator2D:
...
+    .. math::
+        \\int_{x_{\\text{lower}}}^{x_{\\text{upper}}} \\int_{y_{\\text{lower}}(x)}^{y_{\\text{upper}}(x)} f(x, y) \\, dy \\, dx
+        \\approx \\sum_{i=1}^{n_x} \\sum_{j=1}^{n_y} w_{i} w_{j} \\,
+        f\\left( B \\xi_i + A, \\, D(x) \\eta_j + C(x) \\right) \\,
+        B \\cdot D(x)
+    where:
+        -  :math:`x_{\mathrm{lower}}`: Lower limit of integration for the x-dimension.
+        -  :math:`x_{\mathrm{upper}}`: Upper limit of integration for the x-dimension.
+        -  :math:`y_{\mathrm{lower}}(x)`: Lower limit of integration for the y-dimension, a function of :math:`x`.
+        -  :math:`y_{\mathrm{upper}}(x)`: Upper limit of integration for the y-dimension, a function of :math:`x`.
```
jacklovell, line 87: Either double-escape `\mathrm` or (preferably) use a raw string for the docstring (`r'''`) and go with single backslashes for all the Latex code.

cherab/core/math/integrators/integrators2d.pyx
```diff
@@ -59,3 +65,305 @@ cdef class Integrator2D:
...
+
+    @relative_tolerance.setter
+    def relative_tolerance(self, double value):
+
+        if value <= 0:
+            raise ValueError("Relative tolerance must be positive.")
+
+        self._rtol = value
+
+    cdef _build_cache(self):
```
jacklovell, line 244: This doesn't need to be `cdef`, a simple `def` will do.

cherab/core/math/integrators/integrators2d.pyx
```diff
@@ -59,3 +65,305 @@ cdef class Integrator2D:
...
+        -  :math:`D(x) = \\frac{y_{\\mathrm{upper}}(x) - y_{\\mathrm{lower}}(x)}{2}`: Half-width of the y-interval.
+        -  :math:`C(x) = \\frac{y_{\\mathrm{upper}}(x) + y_{\\mathrm{lower}}(x)}{2}`: Midpoint of the y-interval.
+
+    :param Function2D integrand: A 2D function to integrate. Default is `Constant2D(0)`.
+    :param double relative_tolerance: Iteration stops when relative error between
+        last two iterates is less than this value. Default is `1.e-5`.
+    :param int x_max_order: Maximum order on Gaussian quadrature in the x dimension. Default is `50`.
+    :param int x_min_order: Minimum order on Gaussian quadrature in the x dimension. Default is `1`.
+    :param int y_max_order: Maximum order on Gaussian quadrature in the y dimension. Default is `50`.
+    :param int y_min_order: Minimum order on Gaussian quadrature in the y dimension. Default is `1`.
```
jacklovell: I think it's worth clarifying here that `order` is iteratively increased from `min_order` to `max_order` until the desired tolerance is reached, and if either x or y reaches max order before tolerance is reached the other one continues to increase order while the first is pegged at its max order. It takes a lot of reading of the code to realise this.

This explanation is also missing from the 1D case so it would be nice to add it there but at least the 1D integrator's code is simple enough that a reader can quickly work it out.

cherab/core/math/integrators/integrators2d.pyx
```diff
@@ -59,3 +65,305 @@ cdef class Integrator2D:
...
+        Caches the roots and weights of the Gauss-Legendre quadrature.
+        """
+
+        cdef:
+            int order, n, i
+
+        # cache roots and weights for x dimension
+        n = (self._x_max_order + self._x_min_order) * (self._x_max_order - self._x_min_order + 1) // 2
+
+        self._x_roots = np.zeros(n, dtype=np.float64)
```
jacklovell, line 257: These are essentially 1D views of ragged arrays, right? Where each row would be the coefficients for a given order. They would be lower triangular matrices if `min_x_order` = 1. Maybe put a comment somewhere here mentioning this, as it'll save someone at a later date wasting time trying to work out why you're manually indexing instead of just using 2D - or 3D with [x, y, order] - arrays here (like I might have just done...).

cherab/core/math/integrators/integrators2d.pyx
```diff
@@ -59,3 +65,305 @@ cdef class Integrator2D:
...
+
+        i = 0
+        for order in range(self._y_min_order, self._y_max_order + 1):
+            self._y_roots[i:i + order], self._y_weights[i:i + order] = roots_legendre(order)
+            i += order
+
+        self._y_roots_mv = self._y_roots
+        self._y_weights_mv = self._y_weights
+    
+    cpdef double profile_evaluate(self, int n, double x_lower, double x_upper, Function1D y_lower, Function1D y_upper):
```
jacklovell: What's the purpose of this function, and `evaluate_overhead` below? They're not called anywhere: are they left over from testing?

cherab/core/math/integrators/integrators2d.pyx
```diff
@@ -59,3 +65,305 @@ cdef class Integrator2D:
...
+
+        :returns: Gaussian quadrature approximation to integral.
+        """
+
+        cdef:
+            int x_order, y_order
+            int x_ibegin, y_ibegin
+            int i, j
+            double current_integral, previous_integral, y_contribution, error
+            double x, y, x_offset, y_offset, x_slope, y_slope, scale, rtol, y_lower_val, y_upper_val
```
jacklovell: rtol and scale aren't used

cherab/core/math/tests/test_integrators.py
```diff
@@ -67,14 +72,119 @@ def test_properties(self):
...
+        quadrature.integrand = Arg2D("x")
+
+        self.assertEqual(quadrature.relative_tolerance, reltol)
+        self.assertEqual(quadrature.x_min_order, x_min_order)
+        self.assertEqual(quadrature.x_max_order, x_max_order)
+        self.assertEqual(quadrature.y_min_order, y_min_order)
+        self.assertEqual(quadrature.y_max_order, y_max_order)
+        self.assertEqual(quadrature.integrand, Arg2D("x"))
+
+    def test_integrate(self):
```
jacklovell, line 163: Add tests for `x_max_order  > y_max_order` and `x_max_order < y_max_order` to complement the existing `x_max_order == x_min_order`. This may help avoid future regressions.

jacklovell (changes requested):
Very nice, only minor comments from me. Some of the (code comment related) comments apply equally to the existing `GaussianQuadrature` class on which this is based, so if you can add some clarifying code comments there too that would be nice but not a deal breaker.

cherab/core/math/integrators/__init__.pxd
Mateasek: I Agree. I added `GaussianQuadrature1D` and marked `GaussianQuadrature` as deprecated.

cherab/core/math/integrators/__init__.py
Mateasek: done

cherab/core/math/integrators/integrators2d.pxd
Mateasek: true, removed

cherab/core/math/integrators/integrators2d.pyx
Mateasek: Here I went with what @vsnever originally designed for the 1D case and I like this approach. A 2D ragged array would be storing zeros and you would have to track which indexes are valid for the current order anyway. I'm not saying using this flattened case is hugely faster or more efficient, but I think it is as clean as using a 2D matrix for storing the data.

I added explanatory comments for both 1D and 2D cases

cherab/core/math/integrators/integrators2d.pyx
Mateasek: made the docstring a raw string

cherab/core/math/integrators/integrators2d.pyx
Mateasek: Again, here I went with what was done for `GaussianQuadrature1D`. I have no preference here, but I would prefer having it the same way for both 1D and 2D cases. Do you prefer me to change it to a python method in both 1D and 2D classes @jacklovell ?

cherab/core/math/integrators/integrators2d.pyx
Mateasek: I used it to test performance, but I forgot to remove it. Removed...

cherab/core/math/integrators/integrators2d.pyx
Mateasek: The whole evaluate method was refactored and simplified. This should be fixed.

cherab/core/math/integrators/integrators2d.pyx
Mateasek: I added a line explaining the algorithm and expanded the param/ivar lines to give better explanation.

cherab/core/math/tests/test_integrators.py
Mateasek: The test now iterates over max_order product of [30, 40, 50] and min_order product [10, 20, 30].

Mateasek:
Thanks for the review @jacklovell! I implemented changes to take into account the suggestions.

I refactored the `GaussianQuadrature2D.evaluate` method to simplify it. The part of code evaluating the integral was moved to a separate method and the end condicions were simplified. The readability is now improved I think.

cherab/core/math/integrators/integrators2d.pyx
jacklovell: You'll also need to un-escape the backslashes for TeX macros elsewhere in the docstring now that it is a raw string: a single backslash is needed everywhere otherwise the maths isn't rendered correctly when building the HTML docs.

There should also be a blank line after the `.. math::` block and before `where:`, otherwise Sphinx complains with `WARNING: Explicit markup ends without a blank line; unexpected unindent. [docutils]`

`./dev/build_docs.sh` from the root directory of the repo is good for locally checking how the docstrings render. You'll need the `sphinx`, `sphinx-tabs` and `sphinx-rtd-theme` Python packages installed.

cherab/core/math/integrators/integrators2d.pyx
jacklovell: OK, fair enough. For consistency with what's already there `cdef` is fine.

docs/source/math/integrators.rst
```diff
@@ -0,0 +1,11 @@
+
+Integrators
+-------------
+
+.. autoclass:: cherab.core.math.integrators.integrators1d.GaussianQuadrature
```
jacklovell: Should be `GaussianQuadrature1D`

jacklovell:
Great, thanks. 2 final (minor) unresolved comments relating to the documentation and then it'll be good to merge.

docs/source/math/integrators.rst
Mateasek: Good catch, I found few more `GaussianQuadrature` imports/usage and replaced it. I hope that it is all now.

cherab/core/math/integrators/integrators2d.pyx
Mateasek: Thanks, I hope that now it is ok.

Mateasek:
I hope I've addressed all the outstanding comments and problems, so this one is ready for another go.

jacklovell (approved):

Mateasek:
I removed the merge conflicts.

PR #478, opened 2025-02-04
Feature/function6d

### LINKED ISSUE #477: Add scalar Function6D framework to Cherab
Is it worth adding `Function6D` into Cherab? 

`DistributionFunction.evaluate` is close to what a `Function6D.evaluate` would look like in Raysect's function framework. Adding `Function6D` to Cherab would allow us to add an arbitrary `DistributionFunction` implementation initialised with `Function6D` and `Function3D` framework (generalisation of what is now partly done in `Maxwellian`). It would also move us closer to covering non-Maxwellian features in radiation.

I think that implementing it in Cherab would be a good choice, since 6D is too domain specific for Raysect.

I briefly looked into how function framework is implemented in Raysect and it seemed to me the whole procedure would be largely copy-paste job, although I could be mistaken.

I'm opening this issue to initiate some brainstorming and discussion whether it is worth it or not. I would definitely welcome comments from @jacklovell and @vsnever.

Mateasek:
Adds `Function6D` framework inspired by Raysect's `Function3D`. The reason for having it in Cherab is that it is domain specific due to its relation to `DistributionFunction`, as discussed in #477.

The contribution is derived from Raysect's framework and the [PR](https://github.com/raysect/source/pull/426) by @munechika-koyo.


cherab/core/math/function/float/function6d/tests/test_arg.py
```diff
@@ -0,0 +1,54 @@
...
+
+import unittest
+from cherab.core.math.function.float.function6d.arg import Arg6D
+
+# TODO: expand tests to cover the cython interface
+class TestArg6D(unittest.TestCase):
+
+    def test_arg(self):
+        testvals = [-1e10, -7, -0.001, 0.0, 0.00003, 10, 2.3e49]
+        for x in testvals:
```
jacklovell: You can avoid this high level of nesting throughout the tests:

`for (x, y, z, u, w, v) in itertools.product(testvals, repeat=6):`

docs/source/math/function.rst
```diff
@@ -10,6 +10,56 @@ documentation and the Cherab function tutorials.
 Cherab previously provided vector functions which were not present in Raysect.
 New codes should prefer the Raysect vector functions, but the old aliases are preserved for backwards compatibility.
 
+The Function6D framework in Cherab aims to provide a framework for building six-dimensional distribution functions.
+The relation of Function6D to distribution functions makes it domain specific, and so it was included in Cherab's math module.
```
jacklovell: I don't think you need to justify including Function6D here: the first line you've added is sufficient to motivate the framework, and the third line is sufficient to highlight the similarity to Raysect.

jacklovell (commented):
This looks really good. Barring a couple of minor cosmetic changes, I think it's good to go.

docs/source/math/function.rst
Mateasek: I've replaced the nested for cycles with the itertools.product.

Mateasek: I removed the 2nd line

jacklovell (approved):

Issue #294, opened 2021-05-14
Add access to plasma model properties


Mateasek:
Currently PlasmaModel attributes are not accessible from python. Making some of them accessible would provide an easier way of investigating and understanding results provided by cherab. I think in general it would increase the usefulness of the plasma object and sometimes save a lot of time. For example reaction rates could be obtained from the plasma object instead of the need to construct them manually. This could be also useful for debugging of simulations. I propose adding getters to:

ExcitationLine, ExcitationLine:
- _line
- _wavelength
- _rates

TotalRadiatedPower:
- _element
- _charge
- _plt_rate
- _prb_rate

Are there any more objections to this idea? There can be a reason for the current state which I missed. Also are there any more ideas which could be connected to this change?

jacklovell:
I don't see any harm in doing this. You should ensure that the documentation is updated too, to include these properties.

munechika-koyo:
I threw the issue https://github.com/cherab/core/issues/349
I would appreciate it if someone would link this PR to the issue.

Mateasek:
closed with #483

PR #483, opened 2025-06-05
Feature/model attr access

Mateasek:
Added property access  (readonly) to the line and lineshape attributes of emission models, related to #294 . This should allow fully browsing plasma emission models to verify what models were used in observation. This is good for taking logs and result serialisation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Project Changelog
 Release 1.6.0 (TBD)
 -------------------
 
+API changes:
+* Add emission model attribute access to line and lineshape . (#294)
+
 New:
 * Add Function6D framework. (#478)
 * Add e_field attribute to Plasma object for electric field vector. (#465)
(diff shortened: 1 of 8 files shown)


jacklovell:
This looks fine to me. Can you add some simple unit tests to check these attribute accesses? That'll guard against regressions in the future. Then I'm happy to merge.

Mateasek:
Thanks for the review @jacklovell
 
I added tests for plasma models, but the problem is tests fail because the atomic data are not available. There are two options:
1) I remove the tests which require atomic data, which are most of them
2) Atomic data are obtained before tests are initiated, but I have no idea if it is a good idea or even possible

jacklovell:
Add some mock atomic data: doesn't have to be physically meaningful but just the bare minimum for the tests.

Mateasek:
good idea @jacklovell

Mateasek:
I patched the file reading, the tests now pass. I think this is ready to go @jacklovell

Issue #481, opened 2025-04-16
Add GenericDistributionFunction


Mateasek:
The recent addition of the `Function6D` framework allows us to add a generic distribution function defined through the `Function6D`, `Function3D` and `VectorFunction3D` frameworks.

I propose making a new distribution function called `GenericDistributionFunction` which would be defined by `Function6D` giving the phase space particle density (`evaluate`), `Function3D` giving the `density`, and `effective_temperature` and `VectorFunction3D` giving the `bulk_velocity`.

This would allow users to experiment to a great extent without the need to touch cython code.

jacklovell:
Fixed by #494

PR #494, opened 2025-11-25
Feature/generic dist

Mateasek:
Adds the `GenericDistribution` sublcass of `DistributionFunction`, related to #481.

Allows users to build arbitrary particle distribution functions using the Raysect & Cherab function framework.


cherab/core/tests/test_distribution.py
```diff
@@ -0,0 +1,325 @@
...
+
+import unittest
+
+import numpy as np
+
+from cherab.core.distribution import ZeroDistribution, GenericDistribution, Maxwellian
+from raysect.core import Vector3D
+
+ATOMIC_MASS = 1.66053906660e-27
+ELEMENTARY_CHARGE = 1.602176634e-19
```
jacklovell: Single-source these by importing from `cherab.core.utility.constants`. We've been bitten before by "constants" changing, so the fewer places to maintain them the better.

jacklovell (changes requested):
Looks good. Can you fix the merge conflict in the changelog and address one small comment in the test file? Good to see that `ZeroDistribution` is now properly documented too.

cherab/core/tests/test_distribution.py
Mateasek: I agree, but the constants there are cython doubles and I can't import them in python (or am I missing something?). To allow python imports, we would have to change contents of `cherab.core.utility.constants`. It could be done and it is a good idea, but I would prefer doing it as a part of separate issue and PR.

jacklovell: Ah, good point. I'd forgotten they were `cdef`'d and so inaccessible from Python out of the box. It is possible to access them using ctypes, but it's somewhat cumbersome so I agree that belongs in a separate issue/PR.

jacklovell (approved):

cherab/core/tests/test_distribution.py
Mateasek: The constant values are now imported

Mateasek:
I updated the tests to import the constants from cherab. I also fixed one doctring problem and the merge conflict problem. This PR is ready for a new review.

cherab/core/tests/test_distribution.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def test_bulk_velocity(self):
+        """Test that bulk_velocity() returns the correct velocity vector."""
+        # Define velocity function
+        
+        distribution = GenericDistribution(self.phase_space_density, self.density, self.temperature, self.velocity)
+        
+        for x in self.x:
+            for y in self.y:
+                for z in self.z:
```
munechika-koyo:
```suggestion
        for i, j, k in np.ndindex(self.x.size, self.y.size, self.z.size):
        	x, y, z = self.x[i], self.y[j], self.z[k]
```
Just in case, here's another way to avoid excessive nesting in the for loop

Mateasek: Thanks @munechika-koyo, you are right that there are many nested forcycles, some of them with quite extensive depth. I will replace nested for cycles with itertools.product

munechika-koyo: It seems like `itertools.product` is more useful and effective than `numpy.ndindex`.
Thanks for telling me!

demos/particle_distribution/generic_distribution.py
```diff
@@ -0,0 +1,297 @@
...
+bulk_velocity_profile = FloatToVector3DFunction3D(vx_profile, vy_profile, vz_profile)
+
+# initialise the arguments for the 6D function
+x6d, y6d, z6d, vx6d, vy6d, vz6d = (
+    Arg6D("x"),
+    Arg6D("y"),
+    Arg6D("z"),
+    Arg6D("u"),
+    Arg6D("w"),
+    Arg6D("v"),
```
jacklovell, line 105: Am I being ignorant or should these be ordered "u", "v", "w" instead of "u", "w", "v"?

Mateasek: No you are not, but you just pulled a skeleton out of my closet. This is the order I selected for the `Function6D` framework, but I can't imagine why. Since  (x, y, z, u, v, w) is more logical order, I will push a swift fix so we don't accept this into 1.6.

Mateasek: I swapped `v` and `w`.

demos/particle_distribution/generic_distribution.py
```diff
@@ -0,0 +1,297 @@
...
+        
+_, ax = plt.subplots()
+quiver = ax.quiver(X, Y, x_velocity_vals, y_velocity_vals, velocity_magnitude, 
+                   cmap='viridis', scale=1e6, width=0.003)
+plt.colorbar(quiver, ax=ax, label="Velocity Magnitude (m/s)")
+ax.set_xlabel("X (m)")
+ax.set_ylabel("Y (m)")
+ax.set_title("Bulk x, y Velocity Cmponents Field in X-Y Plane (Z=0)")
+ax.set_aspect('equal')
+ax.grid(True)
```
jacklovell, line 297: Add `plt.show()` at the end here so this works when ran as a Python script too.

Mateasek: added

jacklovell (approved):

PR #497, opened 2026-03-22
Fix ADF15 parser and refine regex patterns

munechika-koyo:
## Summary

Because ADF15 raw format for tungsten, like `pec40#w_ic#w12.dat`, is different from what `cherab` has handled like:
```text
C  energy levels
C  -------------
C
C   lv      configuration    (2S+1)L(w-1/2)      energy (cm^-1)
C  ---   ------------------- --------------      --------------
c    1   60964A52B             (1) 0(  0.0)             0.0
c    2   60963A52B51C          (3) 2(  3.0)         16720.4
				:
				:
				:

C  photon emissivity atomic transitions
C  ------------------------------------
C
C   isel  wvlen(A)            transition                type  ispb nspb
C                                                             ispp nspp sz    tg pr wr
C  ----- ---------- ----------------------------------- ----- ---- ---- --    -- -- --
C      1   56.5300  1074(3)1(  2.0)-   2(3)2(  3.0)     excit   1    1  12   569 12  1
C      2   56.5520  1075(1)3(  3.0)-   3(1)4(  4.0)     excit   1    1  12  1068  7  2
C      3   70.1877   414(1)5(  5.0)-   3(1)4(  4.0)     excit   1    1  12   778 31  3
C      4   70.5640   365(3)1(  2.0)-   2(3)2(  3.0)     excit   1    1  12   275 36  4
C      5   70.6699   371(3)3(  4.0)-   3(1)4(  4.0)     excit   1    1  12   756 43  5
C      6   70.7048   367(3)3(  3.0)-   3(1)4(  4.0)     excit   1    1  12   753 44  6
```
I tried to modify the ADF15 parser to match these lines.

## Key changes include:

- Fixed the ADF15 parser to handle additional raw file (tungsten) formats.

- Moved regex pattern definitions outside of loops for efficiency.

### DIFF
--- a/cherab/openadas/parse/adf15.py
+++ b/cherab/openadas/parse/adf15.py
@@ -25,44 +25,44 @@ from cherab.core.utility import RecursiveDict
 from cherab.core.utility.conversion import Cm3ToM3, PerCm3ToPerM3
 
 # Compiled regex patterns for ADF15 file parsing
-_ADF_HEADER_MATCH = re.compile(r'^\s*(\d*) {4}/(.*)/?\s*$')
-_PEC_INDEX_HEADER_MATCH_STANDARD = re.compile(r'^C\s*ISEL\s*(?:WAVELENGTH|WVLEN\(A\))\s*TRANSITION\s*TYPE', re.IGNORECASE)
-_PEC_HYDROGEN_TRANSITION_MATCH = re.compile(r'^C\s*([0-9]*)\.\s*([0-9]*\.[0-9]*)\s*N=\s*([0-9]*) - N=\s*([0-9]*)\s*([A-Z]*)', re.IGNORECASE)
-_PEC_FULL_TRANSITION_MATCH = re.compile(r'^[cC]\s*([0-9]*)\.?\s*([0-9]*\.[0-9]*)\s*([0-9]*)[\(\)\.0-9\s]*-\s*([0-9]*)[\(\)\.0-9\s]*([A-Z]*)', re.IGNORECASE)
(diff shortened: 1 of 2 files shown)


jacklovell:
Looks OK to me. I don't think lifting the declaration of a string literal out of the loop will have any impact on performance, as the assignment operation is trivial. If you really wanted to optimize, you could instead create compiled regex objects using `r = re.compile(expression)` and then use `r.match` instead of `re.match(expression)`, but Python caches the compilation of regexes anyway so unless there's a large number of regular expressions all being used this wouldn't make a difference either. It is more correct to do so if you're refactoring anyway though, so I'd recommend switching to `re.compile`.

The other thing it would be good to have in place are some unit/regression tests for this. We can't use data from real ADF15 files in unit tests for licensing reasons, but I don't see a problem with creating some mock files using the same format but using dummy numerical data to use in unit tests: try dummy variants of Hydrogen, Carbon and Tungsten files.

munechika-koyo:
@jacklovell, I am so sorry for not responding for so long.
I defined some patterns using `re.compile` in the module's global scope and wrote a unit test to parse the ADF15 format.

This is the benchmark result comparing the newly implemented parser and the old one.
```
File: hydrogen
  Main branch (plain re.match):     4.5352s
  Current branch (compiled regex):  2.8475s
  Speedup:                          1.59x
  Improvement:                      +59.3%

File: carbon
  Main branch (plain re.match):     1.9508s
  Current branch (compiled regex):  1.2299s
  Speedup:                          1.59x
  Improvement:                      +58.6%
```

Here is the benchmark script I used.
[benchmark_adf15.py](https://github.com/user-attachments/files/27958869/benchmark_adf15.py)
[old_adf15.py](https://github.com/user-attachments/files/27958870/old_adf15.py)

jacklovell (approved):

PR #503, opened 2026-06-18
Introduce indexed ray transfer APIs and tests

### LINKED ISSUE #310: Add RayTransferEmitter that uses a function to map a point in space to a light source
I would like to add a ray transfer emitter, which would use a custom function to map a point in 3D space to a unit light source in the array. This ray transfer emitter will be compatible with `Discrete2DMesh`, `Discrete3DMesh`, `SOLPSFunction2D`, etc. 

Here is an implementation:
```python
cdef class FunctionRayTransferEmitter(InhomogeneousVolumeEmitter):
    """
    Ray transfer emitter defined with a function, which returns an index of the light source
    for a given point in space.

    :param Function3D raytransfer_function: A function that maps a point in 3D space to
        an array of volumetric light sources with unit emissivity.
        Must return the index of the light source in the array starting with 1.
        Zero means that the point is outside the volume for all light sources
        in the array.
    :param raysect.optical.material.VolumeIntegrator integrator: Volume integrator,
        defaults to `integrator=NumericalIntegrator()`.
    """

    cdef:
        readonly Function3D raytransfer_function

    def __init__(self, raytransfer_function, integrator=None):

        super().__init__(integrator)
        self.raytransfer_function = autowrap_function3d(raytransfer_function)

    @cython.boundscheck(False)
    @cython.wraparound(False)
    @cython.initializedcheck(False)
    cpdef Spectrum emission_function(self, Point3D point, Vector3D direction, Spectrum spectrum,
                                     World world, Ray ray, Primitive primitive,
                                     AffineMatrix3D world_to_local, AffineMatrix3D local_to_world):

        cdef int isource

        isource = <int> libc_round(self.raytransfer_function.evaluate(point.x, point.y, point.z))
        if isource > 0:
            spectrum.samples_mv[isource] += 1.

        return spectrum
```
Usage case:
```python
import numpy as np
from Py3DViewer import Tetmesh
from raysect.optical import World, translate, Point3D, Vector3D, rotate_basis
from raysect.core.math.function.float.function3d.interpolate import Discrete3DMesh
from raysect.optical.observer import PinholeCamera, FullFrameSampler2D
from raysect.primitive import Box
from cherab.tools.raytransfer import FunctionRayTransferEmitter, RayTransferPipeline2D, RoughTitanium

mesh = Tetmesh("path_to_raysect_demos/resources/stanford_bunny.mesh")
try:
    tetra_index = Discrete3DMesh(mesh.vertices, mesh.polys, np.arange(1, mesh.num_polys + 1, dtype=float), False, 0)
    bins = mesh.num_polys
except AttributeError:  # for backward compatibility with older versions of Py3DViewer
    tetra_index = Discrete3DMesh(mesh.vertices, mesh.tets, np.arange(1, mesh.num_tets + 1, dtype=float), False, 0)
    bins = mesh.num_tets

world = World()
lower = Point3D(mesh.bbox[0, 0], mesh.bbox[0, 1], mesh.bbox[0, 2])
upper = Point3D(mesh.bbox[1, 0], mesh.bbox[1, 1], mesh.bbox[1, 2])
emitter = Box(lower, upper, material=FunctionRayTransferEmitter(tetra_index), parent=world)
floor = Box(Point3D(-100, -0.1, -100), Point3D(100, -0.01, 100), world, material=RoughTitanium(0.1))

camera_pos = Point3D(-0.015, 0.204, 0.25)
camera_translate = translate(camera_pos.x, camera_pos.y, camera_pos.z)
camera_rotate = rotate_basis(camera_pos.vector_to(Point3D(0, 0.06, 0)), Vector3D(0, 1, 0))
pipeline = RayTransferPipeline2D()
camera = PinholeCamera((32, 32), parent=world, transform=camera_translate * camera_rotate,
                                   pipelines=[pipeline], frame_sampler=FullFrameSampler2D())
camera.fov = 50
camera.pixel_samples = 1000
camera.spectral_bins = bins
# ray transfer matrix will be calculated for 600.5 nm
camera.min_wavelength = 600.
camera.max_wavelength = 601.

# integration resolution
emitter.material.integrator.step = 0.001

camera.observe()
```
However, for the observers with the large number of pixels, the combination of `FunctionRayTransferEmitter` and `Discrete3(2)DMesh` cannot replace the combination of `Cartesian(Cylindrical)RayTransferEmitter` and `voxel_map` due to significantly lower computational speed.

I will create a PR when #307 is merged to avoid merging conflicts.

munechika-koyo:
## Summary
This PR introduces a general index-function-based ray transfer API by adding indexed emitter and integrator classes, replacing mesh-specific naming with functionality-based naming.

## Key Changes
- Added [cherab/tools/raytransfer/emitters.pyx](cherab/tools/raytransfer/emitters.pyx) implementations:
  - `IndexedRayTransferIntegrator`
  - `IndexedRayTransferEmitter`
- Updated [cherab/tools/raytransfer/emitters.pxd](cherab/tools/raytransfer/emitters.pxd) public Cython declarations for the new APIs.
- Added and updated unit tests in [cherab/tools/tests/test_raytransfer.py](cherab/tools/tests/test_raytransfer.py):
  - evaluate behavior for indexed emission/integration
  - default integrator wiring
  - `Discrete3DMesh`-backed `index_function` behavior
  - equivalent tetrahedral mesh construction checks using `numpy.ndindex` patterns

## Unit Test

| Test | Purpose | Setup | Verification | Why it matters |
|---|---|---|---|---|
| `test_evaluate_function` | Validate that `IndexedRayTransferEmitter` works correctly with `NumericalIntegrator` and maps contributions to the correct bins via `index_function`. | A 3x3x3 Box domain is used with `bins=27`. A diagonal ray crosses the volume. The index function maps in-domain points to 0..26 and returns -1 outside. | Only bins 0, 13, and 26 are non-zero, each with path-length contribution sqrt(3). The output spectrum matches the expected vector with `atol=0.001`. | Confirms correct geometric integration and bin assignment for the index-function-based workflow. |
| `test_default_integrator` | Confirm default integrator behavior when no integrator is explicitly passed. | `IndexedRayTransferEmitter` is created without an integrator argument under the same ray/volume setup as above. | The emitter uses `IndexedRayTransferIntegrator` by default, and the resulting spectrum matches the same expected vector (`atol=0.001`). | Guarantees safe default behavior and avoids mandatory integrator wiring for users. |
| `test_discrete3dmesh_as_index_function` | Validate that `Discrete3DMesh` can be used as an equivalent index function source. | A `Discrete3DMesh` is built from a 4x4x4 vertex grid over 3x3x3 cells; each cube is split into 6 tetrahedra. Cell values follow the same indexing rule as the reference index function. | Representative points across all 27 cells match the reference mapping; outside-domain points return -1; ray-transfer spectra from mesh-based and function-based indexing are equal within `atol=0.001`. | Demonstrates implementation-agnostic design and compatibility with tetrahedral mesh indexing in practical ray-transfer use. |

Executed:

    python -m unittest cherab.tools.tests.test_raytransfer.TestIndexedRayTransferEmitter -v

Result:
- `test_default_integrator`: ok
- `test_discrete3dmesh_as_index_function`: ok
- `test_evaluate_function`: ok
- Ran 3 tests, all passed.

## Example Usage

```python
from raysect.optical import World
from cherab.tools.raytransfer import IndexedRayTransferEmitter

def index_func(x, y, z):
    if x < 0:
        return 0
    return 1

world = World()
material = IndexedRayTransferEmitter(index_func, bins=2)
```

## Compatibility and Risk
- Scope is limited to ray transfer emitter/integrator API naming and related tests.
- Runtime behavior is validated by focused unit tests for indexed evaluation and Discrete3DMesh integration.

## Reviewer Notes
- Main implementation: [cherab/tools/raytransfer/emitters.pyx](cherab/tools/raytransfer/emitters.pyx)
- Public declarations: [cherab/tools/raytransfer/emitters.pxd](cherab/tools/raytransfer/emitters.pxd)
- Validation tests: [cherab/tools/tests/test_raytransfer.py](cherab/tools/tests/test_raytransfer.py)

## Checklist
- [x] API implementation updated
- [x] Cython declaration file updated
- [x] Unit tests added/updated
- [x] Targeted tests executed in pixi test environment

## Benchmark
The appendix in this paper (https://doi.org/10.1063/5.0225703) compared the raytransfer of `Discrete3D` meshes with that of regular grids, showing that the geometry matrix calculation for rectangular grids' raytransfer was much faster than the `Discrete3D` one.

### DIFF
--- a/cherab/tools/raytransfer/emitters.pyx
+++ b/cherab/tools/raytransfer/emitters.pyx
@@ -25,7 +25,7 @@ with other integrators is not guaranteed.
 """
 
 import numpy as np
-from raysect.core.math.function.float.function3d cimport autowrap_function3d
+from raysect.core.math.function.float.function3d cimport autowrap_function3d, Function3D
 from raysect.optical cimport World, Primitive, Ray, Spectrum, Point3D, Vector3D, AffineMatrix3D
 from raysect.optical.material cimport VolumeIntegrator, InhomogeneousVolumeEmitter
 from libc.math cimport sqrt, atan2, M_PI as pi
@@ -263,6 +263,7 @@ cdef class IndexedRayTransferIntegrator(RayTransferIntegrator):
             Vector3D direction
             int isource_pre, isource_current, it, n
             double length, t, dt, x, y, z, res
+            Function3D index_function
 
         if not isinstance(material, IndexedRayTransferEmitter):
             raise TypeError(
@@ -281,6 +282,7 @@ cdef class IndexedRayTransferIntegrator(RayTransferIntegrator):
         dt = length / n  # integration step
         # cython performs checks on attributes of external class,
         # so it's better to do the checks before the loop
+        index_function = material.index_function
         isource_current = -1
         isource_pre = -1
         res = 0
@@ -289,7 +291,7 @@ cdef class IndexedRayTransferIntegrator(RayTransferIntegrator):
             x = start.x + direction.x * t  # x coordinates of the points
             y = start.y + direction.y * t  # y coordinates of the points
             z = start.z + direction.z * t  # z coordinates of the points
-            isource_current = <int>material.index_function(x, y, z)  # get geometry grid index
+            isource_current = <int>index_function(x, y, z)  # get geometry grid index
             if isource_current != isource_pre:  # we moved to the next cell
                 if isource_pre > -1:
                     # writing results for the current source
@@ -655,17 +657,23 @@ cdef class IndexedRayTransferEmitter(InhomogeneousVolumeEmitter):
     """A unit emitter defined by an index function, which can be used
     to calculate ray transfer matrices (geometry matrices).
 
+    This approach is suitable for arbitrary source geometries defined via a callable,
+    for example the 3D application in
+    `K. Munechika et al., Rev. Sci. Instrum. 96, 043509 (2024) <https://doi.org/10.1063/5.0225703>`_.
+
     Note that for performance reason there are no boundary checks in `emission_function()`,
     or in `IndexedRayTransferIntegrator`,
     so this emitter must be placed inside a bounding box.
 
     :param callable index_function: Callable objects taking 3 positional arguments :math:`(X, Y, Z)`.
+        This function should return an integer value; any fractional part will be discarded.
     :param int bins: Number of bins for the spectral array, by default 0.
     :param float integration_step: The length of line integration step, by default 0.01.
     :param VolumeIntegrator integrator: Volume integrator, by default `.IndexedRayTransferIntegrator(integration_step)`.
         Volume integrator, by default `.IndexedRayTransferIntegrator(integration_step)`.
 
     :ivar callable index_function: Callable objects taking 3 positional arguments :math:`(X, Y, Z)`.
+        This function should return an integer value; any fractional part will be discarded.
     :ivar int bins: Number of bins for the spectral array.
 
     .. code-block:: python
--- a/cherab/tools/raytransfer/emitters.pyx
+++ b/cherab/tools/raytransfer/emitters.pyx
@@ -25,6 +25,7 @@ with other integrators is not guaranteed.
 """
 
 import numpy as np
+from raysect.core.math.function.float.function3d cimport Function3D, autowrap_function3d
 from raysect.optical cimport World, Primitive, Ray, Spectrum, Point3D, Vector3D, AffineMatrix3D
 from raysect.optical.material cimport VolumeIntegrator, InhomogeneousVolumeEmitter
 from libc.math cimport sqrt, atan2, M_PI as pi
@@ -216,6 +217,89 @@ cdef class CartesianRayTransferIntegrator(RayTransferIntegrator):
                         spectrum.samples_mv[isource_current] += res  # writing results for the current source
                     isource_current = isource
                     res = 0
+            if isource_current > -1:
+                res += dt
+        if isource_current > -1:
+            spectrum.samples_mv[isource_current] += res
+
+        return spectrum
+
+
+
+cdef class IndexedRayTransferIntegrator(RayTransferIntegrator):
+    """Calculates the distances traveled by the ray through indexed regions in 3D coordinate system: :math:`(X, Y, Z)`.
+
+    This integrator is used with the `IndexedRayTransferEmitter` material class and
+    an index function to calculate ray transfer matrices (geometry matrices).
+    The value for each voxel is stored in respective bin of the spectral array.
+    The distances traveled by the ray through the voxel is calculated
+    approximately and the accuracy depends on the integration step.
+
+    Parameters
+    ----------
+    step : float
+        Integration step, by default 0.001.
+    min_samples : int
+        Number of minimum samples of integration, by default 2.
+    """
+    def __init__(self, double step=0.001, int min_samples=2):
+        super().__init__(step=step, min_samples=min_samples)
+
+    @cython.wraparound(False)
+    @cython.cdivision(True)
+    @cython.initializedcheck(False)
+    @cython.nonecheck(False)
+    cpdef Spectrum integrate(
+        self,
+        Spectrum spectrum,
+        World world,
+        Ray ray,
+        Primitive primitive,
+        InhomogeneousVolumeEmitter material,
+        Point3D start_point,
+        Point3D end_point,
+        AffineMatrix3D world_to_primitive,
+        AffineMatrix3D primitive_to_world
+    ):
+
+        cdef:
+            Point3D start, end
+            Vector3D direction
+            int isource_pre, isource_current, it, n
+            double length, t, dt, x, y, z, res
+
+        if not isinstance(material, IndexedRayTransferEmitter):
+            raise TypeError(
+                'Only IndexedRayTransferEmitter material is supported '
+                'by IndexedRayTransferIntegrator'
+            )
+        start = start_point  # start point in world coordinates
+        end = end_point  # end point in world coordinates
+        direction = start.vector_to(end)  # direction of integration
+        length = direction.get_length()  # integration length
+        if length < 0.1 * self._step:  # return if ray's path is too short
+            return spectrum
+        direction = direction.normalise()  # normalized direction
+        # number of points along ray's trajectory
+        n = max(self._min_samples, <int>(length / self._step))
+        dt = length / n  # integration step
+        # cython performs checks on attributes of external class,
+        # so it's better to do the checks before the loop
+        isource_current = -1
+        isource_pre = -1
+        res = 0
+        for it in range(n):
+            t = (it + 0.5) * dt
+            x = start.x + direction.x * t  # x coordinates of the points
+            y = start.y + direction.y * t  # y coordinates of the points
+            z = start.z + direction.z * t  # z coordinates of the points
+            isource_current = <int>material.index_function(x, y, z)  # get geometry grid index
+            if isource_current != isource_pre:  # we moved to the next cell
+                if isource_pre > -1:
+                    # writing results for the current source
+                    spectrum.samples_mv[isource_pre] += res
+                isource_pre = isource_current
+                res = 0
             if isource_current > -1:
                 res += dt
         if isource_current > -1:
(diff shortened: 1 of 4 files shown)


cherab/tools/raytransfer/emitters.pyx
```diff
@@ -224,6 +225,85 @@ cdef class CartesianRayTransferIntegrator(RayTransferIntegrator):
...
+        direction = start.vector_to(end)  # direction of integration
+        length = direction.get_length()  # integration length
+        if length < 0.1 * self._step:  # return if ray's path is too short
+            return spectrum
+        direction = direction.normalise()  # normalized direction
+        # number of points along ray's trajectory
+        n = max(self._min_samples, <int>(length / self._step))
+        dt = length / n  # integration step
+        # cython performs checks on attributes of external class,
+        # so it's better to do the checks before the loop
```
jacklovell, line 284: I think this comment is redundant, unless you assign `material.index_function` to a local variable outside the loop (like `CartesianRayTransferIntegrator` does, which I assume formed the basis for the new class).

cherab/tools/raytransfer/emitters.pyx
```diff
@@ -569,3 +649,105 @@ cdef class CartesianRayTransferEmitter(RayTransferEmitter):
...
+
+cdef class IndexedRayTransferEmitter(InhomogeneousVolumeEmitter):
+    """A unit emitter defined by an index function, which can be used
+    to calculate ray transfer matrices (geometry matrices).
+
+    Note that for performance reason there are no boundary checks in `emission_function()`,
+    or in `IndexedRayTransferIntegrator`,
+    so this emitter must be placed inside a bounding box.
+
+    :param callable index_function: Callable objects taking 3 positional arguments :math:`(X, Y, Z)`.
```
jacklovell, line 668: You should be clear that this function should return an integer value, as any fractional part will be discarded.

jacklovell (changes requested):
Very cool. If I understand this correctly, it's a generalisation of the existing ray transfer objects where a callable function replaces the voxel map, and this function returns the voxel index at a given point in space. As well as the 3D application in https://doi.org/10.1063/5.0225703 (which you should cite somewhere in the documentation by the way), I can see the possibility of application to axisymmetric voxels of arbitrary poloidal cross section without having to approximate them with a rectangular grid.

I think a demo would be highly beneficial, as it's hard to see a concrete use case from the docstrings alone. Perhaps you could adapt Vlad's Space Invaders demos to showcase the new tools.

cherab/tools/raytransfer/emitters.pyx
munechika-koyo: I assign its attribute into a local valiable `index_function`.

munechika-koyo:
Thank you for your review!
I've been thinking about what makes an effective demo script.
Currently, my ideas are using index functions like:
- analytical geometry with sin, cos, circle, etc.
- using mesh geometry (triangulation for 2D, tetrahedralization for 3D, e.g., Stanford Bunny data used in raysect).

I will try to work out these ideas with AI assistance for now.
Any idea of yours is really helpful.

jacklovell:
Analytically-defined voxels will make a nice demo, agreed. For example, flux-aligned axisymmetric voxels for the generomak equilibrium defined in ($\Delta \psi$, $\Delta \beta$) space for normalised flux $\Psi$ and poloidal angle $\beta$. The `ToroidalVoxelGrid` could handle this but I expect the ray transfer framework to be more performant.

And yes, voxels defined by triangular or tetrahedral meshes would also be a good illustration, showing the ability to extend beyond regular rectilinear grids.

cherab/tools/raytransfer/emitters.pyx
```diff
@@ -569,3 +651,110 @@ cdef class CartesianRayTransferEmitter(RayTransferEmitter):
...
+        object index_function not None,
+        int bins=0,
+        double integration_step=0.01,
+        VolumeIntegrator integrator=None
+    ):
+
+        integrator = integrator or IndexedRayTransferIntegrator(step=integration_step)
+        super().__init__(integrator=integrator)
+
+        self.index_function = autowrap_function3d(index_function)
```
Mateasek, line 725: Making `index_function` a `Function3D` makes the IndexedRay framework specific to isotropic radiation. If you made it `Function6D` which now we have in Cherab, the framework would be generalised to anisotropic applications because you could also pass direction vector components. That is something I'm be very interested in. From the point of view of the code it shouldn't be a large change. What do you think @munechika-koyo?

Mateasek (changes requested):
Hi @munechika-koyo , this is a very nice generalisation of the RT framework! I have one important question to raise. 

Changing the indexing function from `Function3D` to `Function6D` would allow to pass also direction information which would make the new `IRT` framwrok applicable to also anisotropic radiation.

cherab/tools/raytransfer/emitters.pyx
munechika-koyo: That sounds really interesting!
It might also be useful for distributional tomography.
I think we need to add a new API or refactor the existing one to accept `Function6D` as an index function for the `IndexedRayTransferEmitter.emission_function()` and `IndexedRayTransferIntegrator.integrate()` methods.
These changes don't seem straightforward to me, so should we handle them in a separate PR after finalizing this one?

In addition, any good ideas for your demo using `Function6D` would be really helpful.

Mateasek: Maybe I'm missing something, but what else do you need to change except the `Function3D` calls, i.e. in `IndexedRayTransferIntegrator.integrate` you change `<int>index_function(x, y, z)` to `<int>index_function(x, y, z, dx, dy, dz)` and then you need to change the object type and etc.. Maybe I'm missing something.

Mateasek (changes requested):
Thanks @munechika-koyo  for your reaction. It made me look at the Raytransfer framework previously added by @vsnever and the IndexedRaytransfer framework you did in this PR from a broader perspecitve. 

Let me now express how I see the framework and what it should do. I think that RayTransfer in general is there to discretise continuous space into one dimensional integer indexes. Then it uses the integration to add a sensitivity to the integer bins.

What does the current RayTransfer framework does? Its specified by a single aspect:
- The mapping between a cartesian space and a 1D array of integers. There are now two distinct mappings for which two sets of classes were made. That is cylindrical and cartesian mappings.

Now what your contribution does is, and correct me if I'm wrong, that it allows user to define the mapping function. This from my point of view generalises the previous approach. Now, we can have a single set of integrator and emitter, which can be used for the old raytransfer functionality if you provide the right discretising mapping functions.

Now I'm going to propose what I think should be done, but please tell me if I'm missing something and I would also like @jacklovell and @skuba31 to think about this, since it would be a major change:

1.  The `IndexedRaytransfer` approach I described above will replace the older `Raytransfer` because essentially it can do exactly the same thing and we should not be doubling our functionality.
2. The most general indexing function (and replacements for the old framework) together with inverse indexing funcitons will be added to tools to give users the possibility to do some basic and most applications out of the box.
3. Extend the indexing function to accept `Function6D` to make this framework even more general.


I'm sorry to be proposing such a major overhaul of this contribution but there wasn't any previous discussion in an issue, so there was no place to do it. I actually had the approach I described above in my head for a long time and I was very happy to see that you actually did it @munechika-koyo, but I think it could be done in a more general and systematic way which would be actually very powerful. If you think about it, it will allow users to apply discretisation in most of the "information dimensions" Raysect passes to the integrator with the ray which position and direction. This would allow to do anisotropic contribution matrices for example.  Or you could decide that you implement a discretisation function which uses x, y, z and wavelength (and you just don't use 2 parameters in the 6D function evaluation).What do you think?

skuba31:
First of all, great work @munechika-koyo . I did have a look at the proposed changes with the comments by @Mateasek in mind. At this point I think we should consider again whether we want to include these changes into 1.6 at the expense of postponing the release or whether doing the release as soon as possible is the priority. This discussion could potentially take some manpower that could be dedicated to the release.

If not included into 1.6, it could go to development branch and another release be created whenever Matěj finishes his synchrotron model. These two additions would make a logical connection and I think it would justify creation of a release even a short time after 1.6.

I like the idea of having a possibility to define any type of index function including those 6D. A possibility to compute contribution matrices for anisotropic radiation would be quite unique. I agree with Matěj that the functionality of the new classes is more general than those currently implemented.

I am not sure if I understood the proposal by Matěj correctly, but I think the changes could be managed without changing the public API. If I am reading it correctly, the class structure of the changes proposed by Koyo is:
`raysect` emitter/material < base `RayTransfer` < specific `RayTransfer` and `IndexedRayTransfer`.  
But as was pointed out by Matěj, the indexedariants are actually more general than currently implemented ray transfer classes.

I would propose to use the new indexed classes as a new base for raytransfer and specific raytransfer classes would simply have a index function assigned on creation. So the structure would look like:
`raysect` emitter/material < `IndexedRayTransfer` (renamed to simply RayTransfer) < specific `RayTransfer`.  
The individual steps could be
1. Change index_function type from function3d to function6d.
2. Replace old base classes by renaming `IndexedRayTransferEmitter` ->  `RayTransferEmitter` and `IndexedRayTransferIntegrator` -> `RayTransferIntegrator`.
3. Create `cartesian_index_function` and `cylindrical_index_function` both 6d but dropping the direction part of input and use them to define `CartesianRayTransferEmitter` and `CylindricalRayTransferEmitter` both utilizing the new general `(Indexed)RayTransferIntegrator`.
4. Keep `CartesianRayTransferIntegrator` and `CylindricalRayTransferIntegrator` for backward compatibility and possibly add a warning that they are no longer used by the Emitter classes.

If my understanding of the problem is correct, this should include the new functionality while keeping the API basically intact.
Do you think these structural change would be appropriate and acceptable @munechika-koyo @jacklovell @Mateasek ?

Mateasek:
This is exactly what I 

> First of all, great work @munechika-koyo . I did have a look at the proposed changes with the comments by @Mateasek in mind. At this point I think we should consider again whether we want to include these changes into 1.6 at the expense of postponing the release or whether doing the release as soon as possible is the priority. This discussion could potentially take some manpower that could be dedicated to the release.
> 
> If not included into 1.6, it could go to development branch and another release be created whenever Matěj finishes his synchrotron model. These two additions would make a logical connection and I think it would justify creation of a release even a short time after 1.6.
> 
> I like the idea of having a possibility to define any type of index function including those 6D. A possibility to compute contribution matrices for anisotropic radiation would be quite unique. I agree with Matěj that the functionality of the new classes is more general than those currently implemented.
> 
> I am not sure if I understood the proposal by Matěj correctly, but I think the changes could be managed without changing the public API. If I am reading it correctly, the class structure of the changes proposed by Koyo is: `raysect` emitter/material < base `RayTransfer` < specific `RayTransfer` and `IndexedRayTransfer`. But as was pointed out by Matěj, the indexedariants are actually more general than currently implemented ray transfer classes.
> 
> I would propose to use the new indexed classes as a new base for raytransfer and specific raytransfer classes would simply have a index function assigned on creation. So the structure would look like: `raysect` emitter/material < `IndexedRayTransfer` (renamed to simply RayTransfer) < specific `RayTransfer`. The individual steps could be
> 
> 1. Change index_function type from function3d to function6d.
> 2. Replace old base classes by renaming `IndexedRayTransferEmitter` ->  `RayTransferEmitter` and `IndexedRayTransferIntegrator` -> `RayTransferIntegrator`.
> 3. Create `cartesian_index_function` and `cylindrical_index_function` both 6d but dropping the direction part of input and use them to define `CartesianRayTransferEmitter` and `CylindricalRayTransferEmitter` both utilizing the new general `(Indexed)RayTransferIntegrator`.
> 4. Keep `CartesianRayTransferIntegrator` and `CylindricalRayTransferIntegrator` for backward compatibility and possibly add a warning that they are no longer used by the Emitter classes.
> 
> If my understanding of the problem is correct, this should include the new functionality while keeping the API basically intact. Do you think these structural change would be appropriate and acceptable @munechika-koyo @jacklovell @Mateasek ?

Yes, this is exactly  what I had in mind. Whether we include it into 1.6 or not would depend on how long it would take. I personally don't think the changes will be too demanding to implement. Should we create an issue and move this conversation there?

jacklovell:
Discussion about the design should be done in #310

PR #514, opened 2026-08-24
Fix bad import

munechika-koyo:
Fixed the invalid import statement and switched to a relative import.


jacklovell (approved):
Looks fine to me. I think this is a small enough change not to require a changelog entry. Clearly the code is rarely used as no bugs have been reported against the failing import, so user impact should be small.

Issue #507, opened 2026-07-16
Cherab 1.6.0 release


Mateasek:
Tasks and steps before releasing Cherab 1.6.0:

### Pull Requests
- [x] finish code review & merge PR #427 (@jacklovell, @Mateasek, @munechika-koyo, @skuba31 )
- [x] finish code review & merge PR #476 (@jacklovell, @Mateasek, @munechika-koyo, @skuba31 )
- [x] finish code review & merge PR #489 (@jacklovell, @Mateasek, @munechika-koyo, @skuba31 )
- [x] finish code review & merge PR #494 (@jacklovell, @Mateasek, @munechika-koyo, @skuba31 )

### Build
- [ ] Automatic Conda-forge builds triggered by new (release) tag on Cherab repo (@munechika-koyo)
- [x] Develop cibuildwheel workflow for automatically building all release wheels (@jacklovell)

### Documentation
- [x] Change documentation to encourage developers to use `pixi` package manager and recommend it as the primary development tool (@munechika-koyo)
- [ ] check documentation installation with conda (@munechika-koyo)

### Release
- [x] release  Cherab 1.6.0 release candidate (@jacklovell)

### Post-release
- [ ] update cherab-solps repository to require Cherab 1.6.0 (@Mateasek, @skuba31 ) 
- [ ] update cherab-JET repository to require Cherab 1.6.0 (@Mateasek, @skuba31 ) 
- [ ] update cherab-MAST repository to require Cherab 1.6.0 (@jacklovell )

PR #518, opened 2026-08-27
 Add an action to build distributions on tagged releases

jacklovell:
Use the CIbuildwheel Github action to build wheels for all
currently-supported Python versions. Manylinux wheels are produced,
using the manylinux2014 image where possible (to support JET/UKAEA
computers which still run SL7), with manylinux_2_28 for Python 3.14 as
that isn't available in manylinux2014.

Only Linux wheels are built for now: MacOS (and Windows should Raysect
ever support it) are deferred. But the strategy matrix for the wheels
is written in such a way as to easily add other OS's or architectures
later as necessary.

To avoid wasting compute cycles, the workflow is configured to only
run on tagged pushes: we manually tag RC and stable releases so should
only get distributions on these versions. If necessary, tagging a
development version with an M.m.p.dev<n> version can also trigger a
build.

Automatic uploading to PyPI is not implemented: the distribution files
should be downloaded from the `gather_artifacts` job's artifacts and
checked locally before uploading to PyPI. This is to reduce the risk
of uploading broken distributions: if the process proves robust and
reliable then automatic uploading to PyPI can be revisited.

Addresses one of the items in https://github.com/cherab/core/issues/507


.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
+name: Build wheels
+
+# We don't want to spend CI compute cycles on every commit, so restrict building
+# of wheels to releases and candidate releases. These are releases which will be
+# tagged manually with a vM.m.p[.rcN] version.
+# Builds can also be triggered by tagging with a .dev<n> version.
+on:
+  push:
+    tags:
+      - 'v*'
```
munechika-koyo:
```suggestion
  release:
    types:
      - published
```
What about triggering after a release? Seems like it would be more reliable than adding a tag.

munechika-koyo: If we plan to upload artifacts with a GitHub release after adding a tag, it should be okay.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+
+jobs:
+  # Build a wheel for each Python version in a separate parallel job.
+  build_wheels:
+    name: Build wheels for ${{ matrix.py }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]  # Only Linux for now, passive provision for other OS's.
+        python: [cp39, cp310, cp311, cp312, cp313, cp314]
```
munechika-koyo: What about specifying a more detailed matrix?
```suggestion
        python: [cp39, cp310, cp311, cp312, cp313, cp314]
        target:
          - name: Linux (x86_64)
            runner: ubuntu-latest
            platform: manylinux_x86_64
            arch: x86_64

          # TODO: Enable these targets when Raysect supports them.
          # - name: Linux (ARM)
          #   runner: ubuntu-24.04-arm
          #   platform: manylinux_aarch64
          #   arch: aarch64

          # - name: macOS (Apple Silicon)
          #   runner: macos-latest
          #   platform: macosx_arm64
          #   arch: arm64

          # - name: macOS (Intel)
          #   runner: macos-15-intel
          #   platform: macosx_x86_64
          #   arch: x86_64

          # - name: Windows
          #   runner: windows-2022
          #   platform: win_amd64
          #   arch: AMD64

File: .github/workflows/build.yml [hunk 3]
```diff
@@ -0,0 +1,78 @@
...
+# Builds can also be triggered by tagging with a .dev<n> version.
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  # Build a wheel for each Python version in a separate parallel job.
+  build_wheels:
+    name: Build wheels for ${{ matrix.py }} on ${{ matrix.os }}
```
munechika-koyo:
```suggestion
    name: Build wheels for ${{ matrix.python }} on ${{ matrix.target.name }}
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+    runs-on: ${{ matrix.os }}
```
munechika-koyo:
```suggestion
    runs-on: ${{ matrix.target.runner }}
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v4.2.0
+        env:
+          CIBW_BUILD: ${{ matrix.python }}-manylinux_x86_64
```
munechika-koyo:
```suggestion
          CIBW_BUILD: ${{ matrix.python }}-${{ matrix.target.platform }}
          CIBW_ARCHS: ${{ matrix.target.arch }}
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+          # but need newer manylinux_2_28 for Python 3.14
+          CIBW_MANYLINUX_X86_64_IMAGE: >-
+            ${{ contains(matrix.python, 'cp314') && 'manylinux_2_28' || 'manylinux2014' }}
+          # Avoid trying to compile e.g. Pillow (raysect->matplotlib dependency) from
+          # source on older manylinux with newer Python. An old version is fine.
+          CIBW_ENVIRONMENT: 'PIP_PREFER_BINARY=1'
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: cibw-wheels-${{ matrix.os }}-${{ matrix.python }}
```
munechika-koyo:
```suggestion
          name: wheels-${{ matrix.target.platform }}-${{ matrix.python }}
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v4.2.0
+        env:
+          CIBW_BUILD: ${{ matrix.python }}-manylinux_x86_64
+          # Build on manylinux2014 where possible for greatest compatibility,
```
munechika-koyo: Compatibility with glibc lower limit?

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+  # upload to PyPI from this job (since its `needs` entry ensures the SDist and
+  # wheels all built successfully).
+  gather_artifacts:
+    name: Gather SDist and wheel outputs into a single artifact
+    runs-on: ubuntu-latest
+    needs: [build_wheels, make_sdist]
+    steps:
+      - uses: actions/download-artifact@v5
+        with:
+          pattern: cibw*
```
munechika-koyo:
```suggestion
          pattern: wheels-*
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Build sdist
+        run: pipx run build --sdist
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: cibw-sdist
```
munechika-koyo:
```suggestion
          name: sdist
```
because source build isn't related to cibuildwheel.

munechika-koyo (changes requested):

.github/workflows/build.yml
jacklovell: UKAEA is still stuck on SL7 (glibc 2.17, i.e. manylinux2014) and is a significant end user of Cherab so we need to support this.

.github/workflows/build.yml
jacklovell: I had it on `tags` for testing so you can tag a development release and have it built. Seems to work OK, so I'll change it to publishing releases. I didn't want to pollute our releases list while getting the cibuildwheel action to work properly.

.github/workflows/build.yml
jacklovell: With Raysect not providing wheels for anything other than Linux x86-64 (not technically x86 as we don't provide 32-bit wheels) there's only limited usefulness in doing this now. But what you've proposed looks neat enough that it's not the end of the world if it sits here commented out for a while yet.

.github/workflows/build.yml
jacklovell: Also need the sdist, which won't be picked up by this wildcard.

.github/workflows/build.yml
munechika-koyo: Commenting out with a TODO specification should be better for future work.

.github/workflows/build.yml
munechika-koyo: I had missed that.
However, what about disabling this pattern setting?
Are all artifacts downloaded, including both sdist and wheels?

jacklovell: I've changed the prefix from `cibw` to `dist`, so we have `dist-wheels*` and `dist-sdist`. Both SDist and wheels will now match the `dist-*` wildcard pattern.

jacklovell: No other artifacts are currently published, but this'll future-proof against any subsequent runs which may publish artifacts we don't necessarily want in the final gather (e.g. coverage output from tests, or documentation builds).

munechika-koyo (approved):

.github/workflows/build.yml
jacklovell: I've modified this a bit by moving the determination of the manylinux image into a separate step in the job: this allows us to use a bash case statement instead of the Github Actions "ternary" expression so it'll be easier to add support for e.g. cp315 wheels if we decide to do that before dropping manylinux2014 support. Checked locally with [act](https://nektosact.com/) and this does the same thing as the previous `env` expression, i.e. using `manylinux_2_28` for cp314 and `manylinux2014` for the others.

